PR #8, opened 2026-04-29
fix: stop retries piling up in input box and stop redundant resends after Claude resumes

### LINKED ISSUE #1: Foreground is "zsh", not Claude. Skipping send-keys
Using `tmux` or `tmux -CC` with `iTerm` and `ZSH` on `MacOS` I get:

```
[2026-03-31 06:22:59] [INFO] Monitor started for pane %7 (claude PID: 54782)
[2026-03-31 06:23:04] [INFO] Rate limit detected: "⎿  You've hit your limit · resets 4am (America/Buenos_Aires)". Waiting 2276s...

[2026-03-31 07:01:02] [WARN] Foreground is "zsh", not Claude. Skipping send-keys. (Add to foregroundCommands in ~/.claude-auto-retry.json if this is wrong)
[2026-03-31 07:01:32] [WARN] Foreground is "zsh", not Claude. Skipping send-keys. (Add to foregroundCommands in ~/.claude-auto-retry.json if this is wrong)
[2026-03-31 07:02:02] [WARN] Foreground is "zsh", not Claude. Skipping send-keys. (Add to foregroundCommands in ~/.claude-auto-retry.json if this is wrong)
[2026-03-31 07:02:32] [WARN] Foreground is "zsh", not Claude. Skipping send-keys. (Add to foregroundCommands in ~/.claude-auto-retry.json if this is wrong)
[2026-03-31 07:03:02] [WARN] Foreground is "zsh", not Claude. Skipping send-keys. (Add to foregroundCommands in ~/.claude-auto-retry.json if this is wrong)
```

Any idea how to fix this?

ibootz:
## Summary

Three related bugs in the retry pipeline. Each one only becomes visible after the previous is fixed:

1. Retry messages stack in the Claude input box without ever submitting (`sendKeys` race).
2. Once submission works, the same retry is re-sent every 30s for the rest of the window (`isRateLimited` matches stale scrollback).
3. Once the duplicate-send is suppressed, a *new* rate limit later in the day is missed because the monitor reads the lingering "resets HH:MMam" line, computes that as "in the past → +24h", and sleeps until tomorrow.

### 1. tmux send-keys text + Enter race with Claude TUI paste-heuristic

`buildSendKeysArgs` packs the retry text and the `Enter` key into a single `tmux send-keys` invocation. tmux writes them atomically to the pane PTY, so they arrive at the ink-based Claude TUI back-to-back. The TUI's paste-heuristic treats the burst as a paste and the trailing `\r` becomes a paste-newline rather than a submit.

Symptom: 5 identical "Continue where you left off. The previous attempt was rate limited." lines stacked in the input box, none submitted, ending with `Max retries (5) reached`.

Fix: split `sendKeys` into two `tmux send-keys` calls — first `-l <text>` (literal, no key parsing), then a 300 ms delay, then `Enter`. New helpers `buildSendLiteralArgs` / `buildSendEnterArgs`; the legacy `buildSendKeysArgs` is kept and marked `@deprecated` for backward compat.

### 2. Stale rate-limit text in scrollback causes duplicate retries

`isRateLimited()` is stateless: it only checks whether the captured 20 lines contain a `limit` line and a `resets` line within 6 lines of each other. The Claude TUI never clears the rate-limit message from its scrollback, so once the limit text appears it keeps matching forever, even after Claude has resumed.

Symptom (after fix #1): the first retry submission triggers Claude to resume; 30 s later the cooldown ends, the monitor re-captures the pane, the historical rate-limit text is still there, `isRateLimited()` returns true again, and the retry message is sent a second, third, fourth time. Up to 5 redundant submissions per window.

Fix: snapshot a signature of the bottom 5 non-blank lines (`paneSignature`) right before each `sendKeys`. On the next tick, if that signature has changed, treat it as 'Claude is alive again' and exit waiting state immediately, regardless of whether the historical limit text still pattern-matches. Purely additive — `_sigBeforeSend` is null until the first retry, so the first-ever retry path is unchanged.

### 3. Stale "resets" line makes the monitor sleep until next day

After fix #2 the monitor correctly returns to monitoring once Claude resumes — but the rate-limit text still lingers in scrollback. The next time `isRateLimited()` matches it (which can happen on any subsequent tick), `findRateLimitMessage()` was scanning lines forward and returning the *oldest* "resets" line found. If that hour has already passed, `calculateWaitMs()` adds 86_400_000 ms to push it into "tomorrow", and the monitor commits to sleeping ~23–24h. Any *real* fresh rate limit that arrives later the same day is silently missed.

Symptom from a production log: Claude resumed at 03:33 UTC after a real `resets 11:30am Asia/Shanghai` limit; at 03:49 UTC the monitor re-saw the lingering "resets 11:30am" text, computed `waitMs = 85316s` (~23h41m), and slept until the next morning. The user's actual fresh afternoon rate limit was never picked up.

Fix, two independent guards (either one would mitigate, both is robust):
- `findRateLimitMessage` now scans bottom-up. The most recently rendered message wins, so a real new limit at the bottom of the pane is preferred over a stale one above it.
- The monitor's `monitoring → waiting` transition rejects any computed `waitMs > 22h` and stays in monitoring (re-checks each tick). 22h is well above the longest realistic Anthropic reset window and well below the ~24h that gets produced when `+86_400_000` is added to a past-time delta.

## Test plan

- [x] `node --check` passes for `src/tmux.js`, `src/monitor.js`, `src/patterns.js`.
- [x] Manual repro of fix #1: triggered a real rate-limit window, the first retry submitted on its own and Claude resumed.
- [x] Manual repro of fix #2: log shows `Sent retry message (attempt 1)` followed by `User already continued. Attempt counter reset.` instead of attempts 2–5.
- [x] Reproduced fix #3 from a real production log: the same scrollback state that produced `Waiting 85316s` now stays in monitoring until a non-stale message arrives.
- [ ] Add unit tests around `paneSignature`, the `monitoring → waiting` 22h guard, and `findRateLimitMessage` bottom-up scan — happy to do this in a follow-up if a particular shape of test is preferred for this repo.

## Notes

- Three commits kept separate because the bugs are independent at root cause (TUI keystroke timing, monitor state machine, and scrollback-vs-time interaction respectively). Happy to squash if preferred.
- The 300 ms send-keys delay and 22h stale threshold are conservative defaults; both could be made configurable via `~/.claude-auto-retry.json` if there's interest.

### DIFF
--- a/src/monitor.js
+++ b/src/monitor.js
@@ -6,8 +6,19 @@ import { createLogger } from './logger.js';
 
 const DEFAULT_FOREGROUND_COMMANDS = ['node', 'claude', 'npx', 'tsx', 'bun', 'deno'];
 
+// Signature of the bottom of the pane — used to detect whether Claude has
+// actually started responding after we sent a retry message. The rate-limit
+// message lingers in the TUI scrollback even after Claude resumes, so
+// re-running isRateLimited() always returns true and produces redundant
+// retries. A change in the bottom 5 non-blank lines is a much more reliable
+// "Claude is alive again" signal.
+function paneSignature(stripped) {
+  const lines = stripped.split('\n').map(l => l.trimEnd()).filter(l => l.length > 0);
+  return lines.slice(-5).join('\n');
+}
+
 export function createMonitorState() {
-  return { status: 'monitoring', waitUntil: 0, attempts: 0, lastRateLimitMessage: null };
+  return { status: 'monitoring', waitUntil: 0, attempts: 0, lastRateLimitMessage: null, _sigBeforeSend: null };
 }
 
 export async function processOneTick(state, tmuxAdapter, pane, config, isAlive) {
@@ -20,10 +31,22 @@ export async function processOneTick(state, tmuxAdapter, pane, config, isAlive) 
     if (Date.now() < state.waitUntil) return 'waiting';
     if (!isAlive()) return 'exit';
 
+    // After we've sent at least one retry, prefer the "pane moved" signal
+    // over re-pattern-matching the rate-limit text. The limit message stays
+    // in scrollback after Claude resumes, so isRateLimited() would keep
+    // returning true and the loop would resend the retry message every 30s.
+    if (state._sigBeforeSend && paneSignature(stripped) !== state._sigBeforeSend) {
+      state.status = 'monitoring';
+      state.attempts = 0;
+      state._sigBeforeSend = null;
+      return 'user-continued';
+    }
+
     // Always check if rate limit cleared FIRST — even when maxRetries
     // exhausted, the user (or time passing) may have resolved it.
     if (!isRateLimited(stripped, config.customPatterns)) {
       state.status = 'monitoring'; state.attempts = 0;
+      state._sigBeforeSend = null;
       return 'user-continued';
     }
 
@@ -52,6 +75,9 @@ export async function processOneTick(state, tmuxAdapter, pane, config, isAlive) 
 
     // Increment attempts and set cooldown BEFORE sendKeys so that a failure
     // (e.g. pane destroyed) still consumes a retry and avoids tight-loop errors.
+    // Snapshot the pane signature before sending so the next tick can detect
+    // whether Claude actually started responding (vs. our retry being eaten).
+    state._sigBeforeSend = paneSignature(stripped);
     state.attempts++;
     state.waitUntil = Date.now() + 30_000;
     await tmuxAdapter.sendKeys(pane, config.retryMessage);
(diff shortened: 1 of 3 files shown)


cheapestinference:
Thanks @ibootz — two of the three fixes here are now in #22: the send-keys split, and the bottom-up scan so `findRateLimitMessage` parses the newest "resets" line instead of a stale one in scrollback.

I deliberately left out the other two pieces:
- the `waitMs > 22h → skip` guard would break legitimate **weekly-limit** waits, which can legitimately be more than a day away (#15);
- the `paneSignature` "Claude resumed" detection overlaps with existing mitigations (20-line capture window, 30s cooldown, maxRetries cap, and the limit-cleared check).

If you'd like to pursue the `paneSignature` idea on its own, a focused PR would be very welcome. Closing the rest as superseded by #22.

PR #12, opened 2026-05-26
fix 3 issues (issue#7, and 2 more)

SpaceMoehre:
This PR solves 3 issues.

1st issue is that the continue message is not actually sent out to claude and therefore gets written 5 times (Retry-max) to claude code.

2nd issue is that the regex detecting session limits was too restrictive and in my case did not recognize `You've hit your session limit · resets 11:40am (UTC)` 

3rd issue was that claude asks to either stop and wait or upgrade the plan. This pr implements a basic "send Enter" when this is detected to just stop and wait. the session limit message is displayed after selecting to wait

### DIFF
--- a/src/monitor.js
+++ b/src/monitor.js
@@ -1,4 +1,4 @@
-import { stripAnsi, isRateLimited, findRateLimitMessage } from './patterns.js';
+import { stripAnsi, isRateLimited, findRateLimitMessage, isLimitPrompt } from './patterns.js';
 import { parseResetTime, calculateWaitMs } from './time-parser.js';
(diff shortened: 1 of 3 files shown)


cheapestinference:
Thanks @SpaceMoehre — the send-keys fix and the session-limit detection are both in #22. The third part (handling the `/rate-limit-options` menu with a bare Enter) is tracked separately in issue #19 and PRs #17/#14, because the menu's option order varies between Claude Code versions and needs verification before we send Enter blindly. Closing as superseded by #22 for the first two; menu handling continues in #17/#19.

PR #13, opened 2026-06-01
Detect "session limit" and "weekly limit" rate-limit messages

jamesjlopez:
Claude Code shows "You've hit your session limit · resets <time>" and "Weekly limit reached · resets <time>". Neither phrasing was matched by LIMIT_PATTERNS (which expected "your limit", "usage limit", "5-hour limit", etc.), so isRateLimited() returned false and the monitor never waited for the reset or sent the retry — the session stayed paused even after the window reset.

Add /session limit/i and /weekly limit/i to LIMIT_PATTERNS. They still require a nearby "resets" line via the existing co-occurrence check, so they don't false-positive on output that merely mentions the words.

Add test/session-limit.test.js covering both wordings (single-line and multi-line TUI renders) plus negative cases.


cheapestinference:
Thanks @jamesjlopez — "session limit" / "weekly limit" detection is now in #22 (via a broadened qualifier in the existing pattern, keeping the nearby-"resets" co-occurrence check). Closing as superseded by #22.

PR #14, opened 2026-06-08
Fix adjust spend limit menu handling

kkleidal:
This is an extension of #13

#13 only taught claude-auto-retry to recognize newer rate-limit text like “session limit” and “weekly limit.” It improved detection in src/patterns.js, but it still assumed the recovery path was the same as before: wait until reset, then send the normal retry/continue message.

My case was different because Claude didn’t just show a passive rate-limit message. It showed an interactive spend-limit menu with What do you want to do? and Wait for limit to reset. In that flow, sending the usual continue message is not enough. The tool has to drive the TUI first by selecting the Wait for limit to reset option with tmux keypresses, then wait, then send continue. The base branch had no menu detection or tmux navigation logic for that path, so it couldn’t handle this use case.

One follow-up issue I found while adding that support was that generic menu detection can be dangerous: Claude can show other What do you want to do? menus, and stale scrollback can also confuse matching. So the final fix narrows the matcher to the specific spend-limit menu block, rather than any generic menu with similar text.

### DIFF
--- a/src/monitor.js
+++ b/src/monitor.js
@@ -1,13 +1,26 @@
-import { stripAnsi, isRateLimited, findRateLimitMessage } from './patterns.js';
+import { stripAnsi, isRateLimited, findRateLimitMessage, findSpendLimitMenuAction } from './patterns.js';
 import { parseResetTime, calculateWaitMs } from './time-parser.js';
-import { capturePane, sendKeys, getPaneCommand, isProcessForeground } from './tmux.js';
+import { capturePane, sendKeys, sendKeySequence, getPaneCommand, isProcessForeground } from './tmux.js';
 import { loadConfig } from './config.js';
 import { createLogger } from './logger.js';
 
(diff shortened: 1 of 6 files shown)


cheapestinference:
Heads-up @kkleidal — the detection part this builds on (#13's session/weekly wording) is now in #22. The unique value here is the **spend-limit menu navigation**, which overlaps with #17 and issue #19 and is *not* in #22.

Leaving this open as part of the menu cluster. That work still needs verification against the current Claude Code menu (the option order — "Stop and wait" vs "Upgrade your plan" — varies). If you're up for it, a rebase on master once #22 lands would reduce this to just the menu logic and make it easy to review against #17. 🙏

cheapestinference:
Closing — menu/spend-limit handling is now in #26 (merged), with position-independent navigation to "Stop and wait for limit to reset". The session/weekly detection this built on landed earlier in #22. Thank you @kkleidal for surfacing the interactive-menu case. 🙏

Issue #15, opened 2026-06-09
Rate limit not detected: current Claude Code "You've hit your session limit" message slips past all LIMIT_PATTERNS


ccccy666:
## Summary

On the current Claude Code TUI, hitting the 5-hour limit shows:

```
You've hit your session limit · resets 4:50pm (Asia/Shanghai)
```

`isRateLimited()` returns `false` for this string, so the monitor never enters the
`waiting` state, never parses the reset time, and never sends the retry message. The
session just sits there limited. Logs only ever show `Monitor started` / `Claude exited`,
never `Rate limit detected`.

The word **`session`** (between "your" and "limit") is what breaks detection. The tool's
patterns were written against older wordings like `You've hit your limit` /
`Claude usage limit reached`, which still match — but the "session limit" / "weekly limit"
phrasing does not.

## Environment

- `claude-auto-retry` **0.2.2** (global install, Node 24)
- Claude Code TUI (subscription / 5-hour limit)
- tmux path, monitor confirmed alive and polling

## Reproduction

Using the package's own source (no mocking):

```js
// node --input-type=module
import { isRateLimited } from 'claude-auto-retry/src/patterns.js';

console.log(isRateLimited("You've hit your session limit · resets 4:50pm (Asia/Shanghai)")); // false  ← bug
console.log(isRateLimited("You've hit your limit · resets 3pm (Europe/Dublin)"));            // true   (old wording)
console.log(isRateLimited("Claude usage limit reached. Resets at 2pm"));                     // true
```

Also seen with the weekly limit: `You've hit your weekly limit · resets 9am (...)` → `false`.

## Root cause

In `src/patterns.js`, the only limit pattern flexible enough to catch "hit … limit" is:

```js
/(?:hit|exceeded|reached).*(?:your|the)\s*(?:\d+-hour\s+)?limit/i
```

Between `your|the` and `limit` it only allows an optional `\d+-hour`. The real message has
`your session limit` — a plain word in that slot — so it fails, and none of the other six
patterns (`usage limit`, `rate limit`, `\d+-hour limit`, `limit reached`, …) contain
"session"/"weekly" either. Note the *reset* half (`resets 4:50pm (Asia/Shanghai)`) parses
fine on its own; detection requires **both** a limit line and a reset line, so the
unmatched limit line sinks the whole check.

## Proposed fix

Allow an arbitrary qualifier word (session / weekly / usage / N-hour) between `your|the`
and `limit`:

```js
[...]

Issue #10, opened 2026-05-16
Zsh:  syntax error


K-Coconut:
Hi, thanks for your great work. After I installed the plugin, and source the .zshrc, it reported error:

(base) aaaa@SYS-R4000I4:~$ source /home/aaaa/.zshrc
-bash: /home/aaaa/.zshrc: line 21: syntax error near unexpected token `('
-bash: /home/aaaa/.zshrc: line 21: `claude() {'

Appreciate any help !

cheapestinference:
🚀 Fixed in **v0.5.1** — `npm i -g claude-auto-retry@0.5.1`.

The wrapper now unaliases `claude` before defining itself, which fixes the zsh/bash `source` syntax error (#24). Thanks for reporting! Reopen if sourcing still errors. Details: https://github.com/cheapestinference/claude-auto-retry/releases/tag/v0.5.1

Issue #7, opened 2026-04-20
Retry message not sent out


kevineh:
Environment:
Ubuntu 20.04.6 LTS
Claude Code v2.1.42
Tmux 3.0a

Issue:
Read limit error was detected successfully and retry message inputted in Claude Code but not sent out leaving five retry messages in the input box.

Fix:
I was able to make it work by changing this function
`export async function sendKeys(pane, text) {
  // Send text and Enter separately to avoid timing issues where
  // Claude TUI receives them too fast and interprets Enter as newline.
  await execFileAsync('tmux', ['send-keys', '-t', pane, text]);
  await new Promise((r) => setTimeout(r, 50));
  await execFileAsync('tmux', ['send-keys', '-t', pane, 'Enter']);
}`

Not sure if this is just a me problem. Happy to create a pull request to apply the change if needed.

lzkill:
I confirm the issue. Five messages are inputed but not trasmitted to Claude. The sessions only resume after an Enter.

<img width="585" height="135" alt="Image" src="https://github.com/user-attachments/assets/e305cd7c-2a1d-46e0-9d65-3d878ef27e76" />

ZhengyuanCui:
Also have this problem here.

mmarchesotti:
Same problem here

derryltaufik:
same problem here

smprather:
Yep, me too.

smprather:
I know it's only 3 months, but for a showstopper like this (except for @kevineh 's patch), this must be abandonware.

abid-mahdi:
Root cause of the "typed but not sent / only resumes after a manual Enter" behavior: `sendKeys()` sends the message text and `Enter` in a single `tmux send-keys` call. Claude Code's TUI has paste-detection — a burst of characters with a trailing Enter is treated as a *newline inside the input*, not "submit" — so each retry just appends another line (hence the 5 stacked messages).

Fix is to send the text and the Enter as **separate** `send-keys` calls with a short pause between them, so the Enter arrives as its own keypress outside the paste window. The same patch also presses `Escape` first to handle the new interactive `/rate-limit-options` menu (otherwise the Enter confirms "Upgrade your plan"). Full patch + verification in #19.

cheapestinference:
[...]

Issue #6, opened 2026-04-15
Bug: Rate limit reset time is misinterpreted due to timezone, causing 24h delay


OsamuMikumo0704:
### Summary

When using `claude-auto-retry`, the retry wait time is incorrectly calculated due to a timezone mismatch. As a result, the tool waits approximately **one extra day** before retrying, even though the rate limit resets on the same day.

---

### Environment

* OS: Linux (WSL)
* Timezone: Asia/Tokyo
* shell: bash

---

### Steps to Reproduce
1. Hit the Claude rate limit (message shows reset time in Asia/Tokyo)
2. Run `claude-auto-retry status`
---

### Actual Behavior

The log shows a wait time that is about **24 hours longer than expected**, and the retry is scheduled accordingly.

Example:

```
[2026-04-15 09:43:47] [INFO] Rate limit detected: "⎿  You've hit your limit · resets 8pm (Asia/Tokyo)". Waiting 91032s...
```

* `8pm (Asia/Tokyo)` should be later the same day
* However, `Waiting 91032s` ≈ 25 hours, which is incorrect

---

### Expected Behavior

The retry wait time should correctly reflect the reset time on the **same day**, not the next day.

---

### Additional Context

This issue is not only about logs. It also delays continuing.

cheapestinference:
Fix up for review in #23: the timezone convergence now clamps each correction to ±720 min, so offset zones target today's reset instead of tomorrow's (your Asia/Tokyo case included). Will auto-close when #23 merges. Thanks @OsamuMikumo0704.

cheapestinference:
🚀 Fixed in **v0.5.1** — `npm i -g claude-auto-retry@0.5.1`.

The off-by-a-day over-wait in positive/offset timezones is corrected (#23). Thanks for catching this — the 24h delay should be gone now. Reopen if a reset time still computes wrong in your timezone. Details: https://github.com/cheapestinference/claude-auto-retry/releases/tag/v0.5.1

PR #16, opened 2026-06-10
Fix all open upstream issues (#1, #6, #7, #10, #15)

LawfulGremlin:
- #7: Send retry text and Enter as two separate tmux send-keys calls with a 1s delay. When sent in one call, Claude Code's TUI treats the burst as a paste and Enter becomes a newline in the input box instead of submitting. Text is now also sent literally (-l) so key names like "Enter" inside a custom retry message aren't interpreted by tmux.

- #15: Accept qualifier words in the limit pattern so current messages like "You've hit your session limit" / "weekly limit" are detected.

- #6: Include the date when converging on the reset timestamp. For timezones ahead of UTC the naive wall-clock-as-UTC guess lands on the next local day, and matching only hour:minute scheduled the retry ~24h late (e.g. "resets 8pm (Asia/Tokyo)" waited 25h instead of 1h).

- #10: unalias claude before defining the wrapper function. If claude is already an alias (added e.g. by Claude Code's local installer), the function definition is a parse error in zsh. The launcher now also falls back to ~/.claude/local/claude and ~/.local/bin/claude since the alias no longer provides the path.

- #1: Stop skipping retries when pane_current_command reports the login shell (macOS reports "zsh" while Claude runs inside the pane). The monitor now also checks whether the monitored Claude PID is attached to the pane's tty and proceeds if it is.

Adds regression tests for each fix (94 tests passing).

https://claude.ai/code/session_01UFJrNaZXF3d61dT8yp4dMM

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,25 @@ When Claude Code shows *"5-hour limit reached - resets 3pm"*, this tool waits fo
 
 ---
 
+## About this fork
+
+This fork of [cheapestinference/claude-auto-retry](https://github.com/cheapestinference/claude-auto-retry) fixes the open upstream issues:
+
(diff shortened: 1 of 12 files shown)


TianzeWang:
@LawfulGremlin Dude this is super good, why did you delete it? Would you mind open a new fork and uploads your previous branch there? Looks like the original project is no longer maintained..

PR #17, opened 2026-06-10
Handle Claude rate-limit options prompt

MAX-AMMO:
## Summary

Claude Code now sometimes opens `/rate-limit-options` after a subscription/session limit is hit. The existing monitor can detect the limit text and wait/retry, but if Claude is sitting on the interactive options menu, the later retry message may be sent into the wrong UI state.

This patch:
- detects the `/rate-limit-options` prompt when option 1 (`Stop and wait for limit to reset`) is selected
- sends a bare Enter to confirm the wait option
- keeps the monitor in waiting mode so it can send the configured retry message after reset
- recognizes the newer `You've hit your session limit` wording

## Tests

- `npm test`

## Notes

This was reproduced against Claude Code 2.1.170 with output like:

```text
You've hit your session limit · resets 12:10am (Europe/Dublin)
/rate-limit-options
What do you want to do?
❯ 1. Stop and wait for limit to reset
Enter to confirm · Esc to cancel
```

### DIFF
--- a/src/monitor.js
+++ b/src/monitor.js
@@ -1,6 +1,6 @@
 import { stripAnsi, isRateLimited, findRateLimitMessage } from './patterns.js';
 import { parseResetTime, calculateWaitMs } from './time-parser.js';
-import { capturePane, sendKeys, getPaneCommand, isProcessForeground } from './tmux.js';
+import { capturePane, sendKeys, sendEnter, getPaneCommand, isProcessForeground } from './tmux.js';
 import { loadConfig } from './config.js';
 import { createLogger } from './logger.js';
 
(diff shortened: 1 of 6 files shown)


MAX-AMMO:
Follow-up from a live repro: after confirming `/rate-limit-options`, the monitor resumed at reset but repeatedly sent the retry message every 30s because the stale `session limit` text remained visible in pane scrollback.

The second commit changes retry behavior to:
- send literal text and Enter as separate tmux calls
- send one retry message per detected rate-limit event
- wait for the pane to clear stale limit text before retrying again
- ignore stale limit text while Claude is visibly thinking (`Herding` / thinking status)

This matches the failure we saw in Claude Code 2.1.170: the session did continue, but with five duplicate retry prompts in the transcript before the guard was added.

hui-zheng:
I ran into the same issue.

cheapestinference:
Thanks @MAX-AMMO — this is the most complete take on the menu flow, and parsing the reset time directly from the menu text is a nice touch.

One safety concern before merging: `isRateLimitOptionsPrompt` + a bare `sendEnter` assumes "Stop and wait for limit to reset" is the **highlighted** option (it was on your 2.1.170 repro). Issue #19 reports a layout where `❯ 1. Upgrade your plan` is the default — there the same Enter would confirm *Upgrade*. We should make the action cursor-aware (only Enter when `❯` is on "Stop and wait", else navigate) or just Escape-to-dismiss; details and a recommendation in #19.

Also note: the send-keys split and the session/weekly detection in this PR are now in #22 (with a bounded qualifier rather than `.*`). Rebasing on master once #22 lands would reduce this to just the menu logic and make the cursor-aware change easy to review. Keeping this open as the menu-cluster PR.

cheapestinference:
Closing — the menu handling is now in #26 (merged), made **layout-independent**: it locates the `❯` cursor and the "Stop and wait" option and navigates to it regardless of order, so it never confirms "Upgrade your plan" (the concern from #19). Your approach (detect the menu, parse the reset time from it, enter the wait state) is preserved — thank you @MAX-AMMO, it was the basis for the final design. 🙏

PR #18, opened 2026-06-11
fix: update regex to support 'session limit' message

funabiki27:
Anthropic recently updated their rate limit message from 'your 5-hour limit' to 'your session limit' in Claude Code. This PR updates the regex to correctly detect the new message and adds a corresponding test case.


cheapestinference:
Thanks @funabiki27 — the session-limit regex fix is now in #22. I broadened it slightly further (any qualifier word, so "weekly limit" is covered too) and kept a test case like yours. Closing as superseded by #22.

PR #20, opened 2026-06-25
feat: detect sustained API overload (529/500/503) and retry on exponential backoff

romerjon:
## What

Adds an **`overload`** retry path — distinct from the existing usage-limit (subscription rate-limit) path — that detects a *sustained, terminal* API overload error left in the pane (`Overloaded` / `API Error: 529` / `API Error: 500` / `500 Internal server error` / `503`) and retries on a **configurable exponential backoff** with jitter and a cumulative-wait cap.

This is the "new rate-limit patterns" + "notification-adjacent" contribution area from the README, for transient API errors rather than subscription resets.

## Why

Claude Code already retries transient 5xx/529 internally, so this only fires when those internal retries are exhausted and a terminal error sits in the pane — it should rarely trigger. In practice a **`500 Internal server error`** has been observed to halt a session *without* self-resuming, which is why `500` is a default pattern (not opt-in). The mandatory `maxTotalWaitMinutes` cap guards against hammering a genuinely-down endpoint or masking a real outage.

## Design

- **Distinct from usage limits.** New `'overload'` monitor status, separate from `'waiting'`. If both a usage-limit and an overload pattern match, **usage-limit takes precedence** (it has a real reset time; overload is pure backoff).
- **Backoff.** Waits `backoffSeconds[i]` for attempt `i` (default `[30,60,120,240,300]`), then `steadyStateSeconds` (default `300`) once the array is exhausted: `30 → 60 → 120 → 240 → 300 → 300 …`, each with ±`jitterPct`% jitter (default 15, clamped 0–100). Gives up at `maxTotalWaitMinutes` (default 120).
- **Terminal-state gating.** Only retries when Claude is **idle at its prompt** — gated on the existing foreground safety check (`ps -o stat=` `+` flag, with the `pane_current_command` fallback) **plus** a new `isWorking()` check that the `esc to interrupt` working footer is absent. So it never double-drives a session mid internal-retry.
- **Exited-to-shell.** If a `500` ever drops `claude` to the shell, `send-keys` is correctly blocked by the foreground check (never types into bash) and the tool logs `overload-exited-to-shell` rather than masking it. An **opt-in** `relaunchOnExit` (+ `relaunchCommand`, default `claude --continue`) re-enters for anyone who actually observes shell-exits — off by default, since blindly typing into a shell the user may be using is riskier than surfacing the stall.
- **Config-driven + validated.** Partial `overload` blocks shallow-merge onto defaults; bad values fall back (matches the existing validation philosophy): jitter clamped 0–100, positive `backoffSeconds`/`steadyStateSeconds`/`maxTotalWaitMinutes`, non-empty string patterns.

## Config (defaults shown)

```jsonc
"overload": {
  "enabled": true,
  "patterns": ["Overloaded","API Error: 529","529","API Error: 500","500 Internal server error","Internal server error","503","status.claude.com"],
  "backoffSeconds": [30,60,120,240,300],
  "steadyStateSeconds": 300,
  "jitterPct": 15,
  "maxTotalWaitMinutes": 120,
  "retryMessage": "Continue where you left off.",
  "relaunchOnExit": false,
  "relaunchCommand": "claude --continue"
}
```

## Tests

TDD, **+43 new tests** in `test/overload.test.js`: pattern matching (incl. ANSI-wrapped, case-insensitive), config validation/fallbacks, the exact backoff schedule, jitter bounds (±pct, rand extremes), the cap, usage-limit precedence, the gating cases (alive-at-prompt / shell-exit / other-app / working-defer), and that **non-target errors are not retried**. All existing tests stay green.

```
# tests 128
# pass  128
# fail  0
```

**Zero new dependencies.** README updated (new "Overload backoff" section + the sustained-only caveat + the gating rationale; Features list and test count refreshed).

### DIFF
--- a/DESIGN-NOTES.md
+++ b/DESIGN-NOTES.md
@@ -1,0 +1,201 @@
+# Design notes
+
+Forward-looking design notes for `claude-auto-retry`, grounded in a research pass
+(2026-06) against the installed Claude Code binary (v2.1.195, decompiled), the
+Anthropic API error schema, and Claude Code's hooks/transcript surfaces. Ordered by
+leverage. Items marked **[done]** ship in the current version; the rest are proposals
+with enough detail to execute.
+
+## 0. The core problem: we scrape a human-facing render
+
+The overload path detects a *terminal* API error by scraping `tmux capture-pane` and
+string-matching. That is fundamentally a heuristic over output meant for humans, and
+every false positive we've hit (a `res.status(503)` under edit, an HTTP status code in
+tool output in scrollback) and every false negative we worry about is a symptom of that
+layer choice.
+The hardening below makes the scraper as good as a scraper can be — but the strategic
+move is to stop using the scrape as the *trigger*.
+
+### Verified ground truth (informs everything else)
+
+- **Two render forms, opposite meaning.** Terminal (retries exhausted):
+  `API Error: <code> <body>` (colon form). Transient (still retrying):
+  `API Error (<code> …) · Retrying in 5s · attempt 3/10` (parens, and/or a `· Retrying`
+  suffix on the colon form). Acting on the transient form interrupts Claude's *own*
+  backoff. **[done]** — we require the colon form and suppress the retry suffix via the
+  working gate.
+- **Error → HTTP status → retryability.** Retryable: `429 rate_limit_error`,
+  `500 api_error`, `504 timeout_error`, `529 overloaded_error`, plus edge `502/503`
+  (no JSON body — plain text/HTML from Cloudflare/envoy, e.g. `503 no healthy upstream`).
+  Never retry: `400/401/402/403/404/413`. The official SDK retry set is `408/409/429 +
+  all 5xx`. **[done]** — default patterns cover `429|500|502|503|504|529` + `overloaded_error`.
+- **API-429 has no 3-digit code in the slot.** It renders
+  `API Error: Server is temporarily limiting requests (not your usage limit) · Rate
+  limited`. A code-keyed matcher misses it. **[done]** — phrase pattern added.
+- **529 is global capacity, not per-key.** Naive fixed backoff makes every client
+  resynchronize into the same overload window → use **full jitter**. See §3.
+- **Streaming caveat.** An error after a 200 (SSE `error` event) bypasses HTTP status.
+  Claude streams, so a mid-response overload may render differently than a pre-flight
+  failure. Not currently special-cased; low observed frequency.
+
+## 1. Replace the trigger with `StopFailure` (highest leverage) — **[done, v0.4.0]**
+
+**Implemented.** The launcher stamps `CLAUDE_AUTO_RETRY_PANE` onto claude's env; the
+`StopFailure` hook (`claude-auto-retry _stopfailure-hook`, installed via
+`install-hook`) runs as a claude child, inherits that var, and writes a pane-keyed
+marker under `~/.claude-auto-retry/events/` for retryable error types
+(`overloaded|server_error|rate_limit`). The monitor reads the marker for its own pane
+each tick; the first marker latches `eventMode`, which disables the scraper for that
+session. Edge-triggered: one backoff-then-send per failure, then back to monitoring;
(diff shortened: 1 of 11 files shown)


romerjon:
Pushed a follow-up commit that hardens this feature and adds an optional event-driven mode:

**1. Fixes false positives in the scraper.** The original patterns matched bare substrings (`529`, `503`, `500 Internal server error`, `status.claude.com`) across the whole pane capture, so ordinary content — `res.status(503)` under edit, an HTTP code in scrollback, a `status.claude.com` link in a comment — looked identical to a real terminal error and injected spurious retries. Detection is now:
- anchored to Claude Code's actual **colon-form** render `API Error: <code>` for the retryable set (`429|500|502|503|504|529`) + the `overloaded_error` JSON type + the dedicated API-429 phrase — no bare status numbers;
- **terminal-vs-transient aware** — requires the colon form and suppresses the `· Retrying… / attempt n/m` suffix, so it never interrupts Claude's own internal backoff;
- tail-scoped (last 12 lines) so scrollback can't trigger or suppress it;
- logged with the matched pattern + offending line.

**2. Adds opt-in event-driven detection** via the `StopFailure` hook (`install-hook`), which fires on Claude Code's typed API-error event — no scraping, no ambiguity. The scraper stays as the fallback when the hook isn't installed. See `DESIGN-NOTES.md`.

Tests 133 → 158. Also sets `package.json` `repository.url` back to this upstream repo. Happy to split the bug-fix and the feature into separate PRs if you'd prefer to review them independently.

cheapestinference:
Reviewed thoroughly, including running the full suite locally: **158/158 tests pass**. This is excellent work, @romerjon — easily the most rigorous PR on the tracker.

**What stands out**
- Clean separation of the overload path (seconds-scale backoff) from the usage-limit path (hours-scale reset); usage-limit correctly takes precedence.
- Dual-channel design: the authoritative `StopFailure` hook (verified against the v2.1.195 binary in DESIGN-NOTES) with a scraper fallback, and `eventMode` latching so the scraper is disabled once the hook proves live.
- Strong false-positive guards — anchored patterns (no bare status codes), tail-only inspection, and the `isWorking()` gate so we never interrupt Claude's own internal 5xx/529 retries. The negative test cases (`res.status(503)`, status codes in comments, the transient parens form) directly kill the historical false-"continue" bug.
- Conservative defaults: `relaunchOnExit: false` (never types into a shell the user might be using), a cumulative `maxTotalWaitMinutes` cap that gives up loudly, jitter, and pure testable helpers with injectable `rand`.
- Config validation clamps/repairs every field and drops uncompilable regexes.

**Notes before merge (none blocking):**
1. **Rebase on #22** — both touch `patterns.js` (in different regions: #22 changes `LIMIT_PATTERNS`/`findRateLimitMessage`, this adds new functions). Likely auto-mergeable, but a rebase makes it clean. Merging #22 → #23 → #24 first, then this, is the smooth order.
2. **`overload.enabled: true` by default** is the one real decision for the maintainer: it turns this on for everyone. Given the guards I think on-by-default is defensible (a terminal 500 can halt a session with no self-recovery), but flagging it explicitly.
3. Minor: `applyJitter` uses `Math.random` by default — fine for the daemon; tests already inject `rand`.

**Recommendation:** approve, merge after #22/#23/#24 with a rebase. The decision on enabled-by-default is yours; everything else is solid. 🙏

cheapestinference:
✅ **Integrated and shipped** in #27 (merged to master). Your commits were brought over with authorship intact; I only resolved the conflicts with the menu handler (#26) and the send-keys/detection work (#22) that merged in the meantime — **no logic changes**. Full suite is green at 187/187 with your overload/events/config tests included. Thank you for an exceptionally well-engineered contribution, @romerjon — the false-positive guards and the StopFailure-hook design were a cut above. 🙏

cheapestinference:
Thanks again @romerjon — your overload-backoff work was integrated via #27 and is now shipped in **v0.5.1** (`npm i -g claude-auto-retry@0.5.1`). 🙏 https://github.com/cheapestinference/claude-auto-retry/releases/tag/v0.5.1

PR #22, opened 2026-06-29
fix: reliable retry submission + session/weekly limit detection (#7, #15)

### LINKED ISSUE #7: Retry message not sent out

### LINKED ISSUE #15: Rate limit not detected: current Claude Code "You've hit your session limit" message slips past all LIMIT_PATTERNS

cheapestinference:
## What

Fixes the two most-reported bugs in the tracker, each with tests:

### 1. Retry message never submitted (#7) — the "showstopper"
`sendKeys()` sent the retry text and `Enter` in a single `tmux send-keys`
call. Claude Code's Ink TUI treats that burst as a bracketed paste, so the
trailing Enter becomes a newline and the message piles up in the input box
unsent (the "5 messages, never sent" reports). Now split into **literal
text (`-l`) → short pause → bare Enter**, so the Enter lands outside the
paste window. `-l` also stops tmux key names inside a custom `retryMessage`
(e.g. "Enter", "C-c") from being interpreted as keypresses.

### 2. "session limit" / "weekly limit" not detected (#15)
Anthropic changed the wording to `You've hit your session limit` /
`weekly limit`. The limit regex only allowed an optional `N-hour` qualifier,
so these slipped past `isRateLimited()` and the monitor never waited or
retried. Broadened to accept up to three qualifier words (still backward
compatible with `5-hour limit` and bare `your limit`).

Also: `findRateLimitMessage()` now scans **bottom-up**, so it parses the
newest `resets` line instead of a stale one lingering in TUI scrollback.

## Closes
- Closes #7
- Closes #15

## Relationship to existing PRs
Consolidates the overlapping community fixes into one tested change:
- send-keys fix: supersedes #11, and the send-keys portion of #8, #12, #16
- detection fix: supersedes #18, #13, and the regex portion of #8, #12, #14

Deliberately **not** included (need separate review / out of scope here):
- `/rate-limit-options` menu handling (#17/#19) — option order varies, needs
  verification against current Claude Code before driving the menu
- #8's `waitMs > 22h → skip` guard — would break legitimate weekly-limit waits

## Tests
`npm test` → 93/93 pass (4 new: literal send-keys args, session/weekly
wording, no-qualifier regression, newest-resets-line).

### DIFF
--- a/src/patterns.js
+++ b/src/patterns.js
@@ -23,7 +23,7 @@ export function stripAnsi(text) {
 // Detection: find a "limit" line and a "resets" line within 6 lines of each other.
 
 const LIMIT_PATTERNS = [
-  /(?:hit|exceeded|reached).*(?:your|the)\s*(?:\d+-hour\s+)?limit/i,  // "hit/exceeded/reached your limit"
+  /(?:hit|exceeded|reached).*(?:your|the)\s*(?:[\w-]+\s+){0,3}limit/i,  // "hit/exceeded/reached your [session|weekly|5-hour] limit"
   /\d+-hour limit/i,                                // "5-hour limit"
   /limit reached/i,                                  // "limit reached"
   /usage limit/i,                                    // "usage limit"
@@ -73,14 +73,17 @@ export function isRateLimited(text, customPatterns = []) {
 export function findRateLimitMessage(text, customPatterns = []) {
   const lines = stripAnsi(text).split('\n');
 
-  // Return the "resets" line — that's what parseResetTime needs
-  for (const line of lines) {
-    if (RESET_PATTERNS.some(p => p.test(line))) return line.trim();
+  // Scan from the bottom up — the most recent "resets" line is the one to
+  // parse. The Claude TUI never clears earlier rate-limit messages from
+  // scrollback, so a forward scan would lock onto a stale line (e.g. an old
+  // "resets 11:30am" lingering above a fresh "resets 4:30pm").
+  for (let i = lines.length - 1; i >= 0; i--) {
+    if (RESET_PATTERNS.some(p => p.test(lines[i]))) return lines[i].trim();
   }
 
-  // Fallback: any "limit" line
-  for (const line of lines) {
-    if (LIMIT_PATTERNS.some(p => p.test(line))) return line.trim();
+  // Fallback: any "limit" line, also scanned from the bottom.
+  for (let i = lines.length - 1; i >= 0; i--) {
+    if (LIMIT_PATTERNS.some(p => p.test(lines[i]))) return lines[i].trim();
   }
 
   return null;
--- a/src/tmux.js
+++ b/src/tmux.js
@@ -7,8 +7,32 @@ export function buildCaptureArgs(pane, lines = 200) {
   return ['capture-pane', '-t', pane, '-p', '-S', `-${lines}`];
 }
 
+// Submit delay: when sending a message to a TUI like Claude Code (Ink-based),
+// the text and the submitting Enter must be sent as TWO separate tmux send-keys
+// calls with a brief pause between them. Without the pause, Ink on Linux often:
+//   - interprets the Enter as a newline within a bracketed-paste burst and just
+//     inserts "\n" into the input box instead of submitting (most common case)
+//   - or processes the Enter before React reconciliation has incorporated the
+//     text into input state, dropping the submit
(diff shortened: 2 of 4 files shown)


cheapestinference:
✅ Shipped in **v0.5.1** — `npm i -g claude-auto-retry@0.5.1`. Reliable retry submission + session/weekly limit detection are live. Thanks! 🙏 https://github.com/cheapestinference/claude-auto-retry/releases/tag/v0.5.1

PR #4, opened 2026-04-14
fix: correct 24h over-wait for UTC+ timezone rate limit resets

kpmrozowski:
## Summary

- For UTC+ timezones (e.g. `Europe/Warsaw` = UTC+2), `getTargetTimestamp` in `time-parser.js` converges on **tomorrow's** occurrence instead of today's, causing a ~26h wait instead of ~2h
- Root cause: line 66 treats the target wall-clock time as UTC (`new Date(targetStr + 'Z')`). For `Europe/Warsaw`, `"22:00Z"` = `00:00 Warsaw` (next day). The iterative correction then adds 22h forward, landing on tomorrow's 10pm
- Fix: after convergence, roll back 24h if today's occurrence is still in the future
- Added two regression tests covering the fix and the "already passed" case

### Reproduction

```
[2026-04-14 19:49:16] [INFO] Rate limit detected: "⎿  You've hit your limit · resets 10pm (Europe/Warsaw)". Waiting 87104s...
```

Expected: ~7900s (~2h 11m). Actual: 87104s (~24h 12m).

## Test plan

- [x] Existing 18 tests pass unchanged
- [x] New test: `"resets 10pm (Europe/Warsaw)"` at 19:49 Warsaw → wait ~2h (not ~26h)
- [x] New test: same timezone after reset passed (22:30 Warsaw) → correctly waits ~23.5h for tomorrow
- [ ] Verify with other UTC+ zones (Asia/Kolkata UTC+5:30, Asia/Tokyo UTC+9)
- [ ] Verify UTC- zones (America/New_York) remain unaffected

PR #23, opened 2026-06-29
fix: correct off-by-a-day retry wait in offset timezones (#6)

### LINKED ISSUE #6: Bug: Rate limit reset time is misinterpreted due to timezone, causing 24h delay

cheapestinference:
## What

In timezones with a non-zero UTC offset, `calculateWaitMs()` could schedule
the retry **~24h late** — long after the user has left the session.

## Root cause

`getTargetTimestamp()` converges on the reset time by formatting a guess in
the target zone and applying the raw `hour:minute` delta. In a UTC+ zone the
naive UTC guess for "today HH:MM" already falls *past midnight* in the target
zone, so the correction is a near-full-day jump that lands on **tomorrow's**
occurrence:

- `resets 11:40pm (Australia/Melbourne)` → waited ~37.6h instead of ~13.6h
- `resets 10pm (Europe/Warsaw)` → waited ~26h instead of ~2h

## Fix

Normalize each iteration's correction to `[-720, +720]` minutes, so it always
takes the minimum-magnitude adjustment and stays anchored on today's local
date. Symmetric for positive- and negative-offset zones.

## Closes
- Closes #6

## Relationship to existing PRs
- Supersedes **#4** (same root cause; #4 rolled back a day *after* convergence
  on the non-ambiguous path — this fixes the convergence math itself, so the
  ambiguous-hour path benefits too).

## Tests
`npm test` → 96/96 pass (3 new regressions: positive-offset same-day,
negative-offset same-day, and already-passed → tomorrow).

### DIFF
--- a/src/time-parser.js
+++ b/src/time-parser.js
@@ -76,7 +76,13 @@ export function calculateWaitMs(parsed, marginSeconds = 60, fallbackHours = 5, n
       const ch = parseInt(fp.find(p => p.type === 'hour').value) % 24;
       const cm = parseInt(fp.find(p => p.type === 'minute').value);
 
-      const diffMin = (h - ch) * 60 + (m - cm);
+      // Normalize to [-720, +720] minutes so we take the minimum-magnitude
+      // correction. Otherwise, in a UTC+10 tz looking for 23:40, the naive UTC
+      // guess formats as 09:40 next day local, and a raw +14h adjustment lands
+      // on tomorrow's occurrence instead of today's (the off-by-a-day bug).
+      let diffMin = (h - ch) * 60 + (m - cm);
+      diffMin = ((diffMin % 1440) + 1440) % 1440;
+      if (diffMin > 720) diffMin -= 1440;
       if (diffMin === 0) break;
       candidate += diffMin * 60_000;
     }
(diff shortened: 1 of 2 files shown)


cheapestinference:
✅ Shipped in **v0.5.1** — `npm i -g claude-auto-retry@0.5.1`. The offset-timezone off-by-a-day fix is live. Thanks! 🙏 https://github.com/cheapestinference/claude-auto-retry/releases/tag/v0.5.1

PR #24, opened 2026-06-29
fix: unalias claude before wrapper definition to fix shell source error (#10)

### LINKED ISSUE #10: Zsh:  syntax error

cheapestinference:
## What

Sourcing the rc file fails with:

```
.zshrc: line 21: syntax error near unexpected token '('
.zshrc: line 21: `claude() {'
```

## Root cause

When `claude` is already a shell **alias** (Claude Code's own local installer
adds `alias claude=...`), the shell expands that alias while parsing our
`claude() {` function definition, producing the malformed token and the syntax
error. The rc file then aborts mid-source.

## Fix

Prepend `unalias claude 2>/dev/null || true` inside the managed block, so the
alias is removed before the function definition is parsed. Harmless when no
alias exists.

## Verification

Reproduced with bash (`shopt -s expand_aliases` to mimic an interactive shell):

```
# before: alias claude=... ; claude() { ... }
syntax error near unexpected token '('     ← the reported bug

# after: alias claude=... ; unalias claude 2>/dev/null || true ; claude() { ... }
REACHED-END                                ← clean
```

## Closes
- Closes #10

## Tests
`npm test` → 86/86 pass (1 new: asserts the injected wrapper places `unalias
claude` before the function definition).

### DIFF
--- a/src/wrapper.sh
+++ b/src/wrapper.sh
@@ -1,4 +1,9 @@
 # >>> claude-auto-retry >>>
+# Drop any pre-existing `claude` alias (Claude Code's own installer adds one)
+# before defining the wrapper function. Without this, the shell expands the
+# alias while parsing `claude() {`, producing "syntax error near unexpected
+# token '('" when the rc file is sourced.
+unalias claude 2>/dev/null || true
 claude() {
   if [ "${CLAUDE_AUTO_RETRY_ACTIVE}" = "1" ]; then
     command claude "$@"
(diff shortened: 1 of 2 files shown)


cheapestinference:
✅ Shipped in **v0.5.1** — `npm i -g claude-auto-retry@0.5.1`. The wrapper unalias fix for the zsh/bash source error is live. Thanks! 🙏 https://github.com/cheapestinference/claude-auto-retry/releases/tag/v0.5.1

PR #26, opened 2026-06-29
fix: drive /rate-limit-options menu to "Stop and wait", any layout (#19)

### LINKED ISSUE #19: Recovery types into the new /rate-limit-options menu and confirms "Upgrade your plan" instead of resuming
## Summary

On current Claude Code, hitting a session/weekly limit now shows an **interactive menu** instead of a plain prompt:

```
> /rate-limit-options
You've hit your session limit · resets 6:50pm (Europe/London)
What do you want to do?
❯ 1. Upgrade your plan
  2. Stop and wait for limit to reset
```

`sendKeys()` runs `tmux send-keys -t <pane> "<retryMessage>" Enter`. Against this menu:

- the typed message is ignored (it's a selector, not a text box), and
- the trailing **Enter confirms the highlighted default — "1. Upgrade your plan"** — instead of resuming work.

This is distinct from #15 (detecting the "session limit" wording) and compounds #7 (message not submitted): even after detection and submit are fixed, the menu still swallows the Enter.

## Environment

- `claude-auto-retry` 0.2.2 (global install), Node 24, macOS, tmux 3.6b
- Claude Code TUI, subscription session limit

## Root cause (two issues, same code path)

1. **Menu**: a single Enter into the selector confirms the default option ("Upgrade your plan").
2. **Submit** (overlaps #7): sending text and `Enter` in one `send-keys` call makes Claude Code's paste-detection treat the trailing Enter as a *newline*, so the message lands in the input box unsent and piles up one line per retry (the "5 messages, never sent" reports in #7).

## Fix

Dismiss any menu first (`Escape`), type the text, pause, then send a **separate** `Enter` (outside the paste window):

```js
// src/tmux.js
const delay = (ms) => new Promise(resolve => setTimeout(resolve, ms));

export async function sendKeys(pane, text) {
  // 1. Dismiss any interactive prompt (e.g. /rate-limit-options) so the later
  //    Enter resumes work instead of confirming "Upgrade your plan". Harmless
  //    at a normal idle prompt.
  await execFileAsync('tmux', ['send-keys', '-t', pane, 'Escape']);
  await delay(250);
  // 2. Type the message WITHOUT Enter. Combining text + Enter in one send-keys
  //    call trips Claude Code's paste-detection -> the Enter becomes a newline
  //    and the message is never submitted (see #7).
  await execFileAsync('tmux', ['send-keys', '-t', pane, text]);
  // 3. Submit with a SEPARATE Enter after a pause, outside the paste window.
  await delay(400);
  await execFileAsync('tmux', ['send-keys', '-t', pane, 'Enter']);
}
```

## Verification

Reproduced both failure modes and validated the fix against a tmux mock of the menu plus a prompt that models Claude's paste-detection (a lone Enter submits; text+Enter inserts a newline):

- unpatched + menu present → confirms "Upgrade your plan" ❌
- unpatched + plain prompt → message piles up unsent (matches #7) ❌
- patched (both cases) → menu cancelled / message submitted, session resumes ✅

I also exercised the state machine (`processOneTick`) and reset-time parsing with the named-timezone `(Europe/London)` and AM-crossing-midnight cases — those are fine.

One note for maintainers: there's no supported Claude Code setting to disable this menu, so the wrapper has to dismiss it itself. Happy to send a PR with the above plus tests.

cheapestinference:
## What

When a session/weekly limit is hit, newer Claude Code shows an interactive menu:

```
What do you want to do?
❯ 1. Upgrade your plan
  2. Stop and wait for limit to reset
```

A bare Enter confirms the **highlighted default**. The option order varies by
version — #17 saw "Stop and wait" highlighted; #19 saw "Upgrade your plan" — so
pressing Enter blindly can confirm "Upgrade".

## Fix (position-independent)

- `isRateLimitOptionsPrompt()` detects the menu.
- `menuStepsToWaitOption()` locates the cursor (`❯`) and the "Stop and wait"
  option among the numbered lines and returns the signed number of cursor moves.
- The monitor sends that many `Down`/`Up` keys, then `Enter`, then enters the
  normal hours-scale wait (reset time parsed from the menu text itself).
- If the layout can't be read (no cursor), it **refuses to press Enter** and
  rechecks — never risks confirming "Upgrade".

Works for both observed layouts and is robust to future reordering.

## Closes
- Closes #19

## Supersedes
- #17 and #14 (menu handling) — same goal, made layout-robust and rebased on the
  already-merged send-keys (#22) work.

## Tests
`npm test` → 114/114 pass. New: detection through both layouts + ANSI, navigation
steps (+1 / 0 / -1 / null), and monitor-level tests asserting the exact key
sequence (`Down,Enter` / `Enter`) and the no-cursor refusal.

### DIFF
--- a/src/monitor.js
+++ b/src/monitor.js
@@ -1,6 +1,6 @@
-import { stripAnsi, isRateLimited, findRateLimitMessage } from './patterns.js';
+import { stripAnsi, isRateLimited, findRateLimitMessage, isRateLimitOptionsPrompt, menuStepsToWaitOption } from './patterns.js';
 import { parseResetTime, calculateWaitMs } from './time-parser.js';
-import { capturePane, sendKeys, getPaneCommand, isProcessForeground } from './tmux.js';
+import { capturePane, sendKeys, sendKey, getPaneCommand, isProcessForeground } from './tmux.js';
 import { loadConfig } from './config.js';
 import { createLogger } from './logger.js';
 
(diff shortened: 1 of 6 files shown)


cheapestinference:
✅ Shipped in **v0.5.1** — `npm i -g claude-auto-retry@0.5.1`. `/rate-limit-options` menu navigation to "Stop and wait" (any layout) is live. Thanks! 🙏 https://github.com/cheapestinference/claude-auto-retry/releases/tag/v0.5.1

PR #28, opened 2026-06-29
fix(monitor): foreground-guard the /rate-limit-options menu handler

cheapestinference:
The menu handler (#26) sent menu keys without the foreground safety check the retry/overload paths use. If the user focused another app in the pane while the menu was up, we'd drive that app. Gate menu navigation behind `checkForeground()`; skip (`skipped-not-claude`) with a cooldown when Claude isn't foreground.

Found during runtime verification. `npm test` → 188/188 (1 new test).


cheapestinference:
✅ Shipped in **v0.5.1** — `npm i -g claude-auto-retry@0.5.1`. The foreground guard on the menu handler is live. Thanks! 🙏 https://github.com/cheapestinference/claude-auto-retry/releases/tag/v0.5.1

PR #29, opened 2026-06-29
fix(monitor): stop usage-retry from spamming a resumed/working session

romerjon:
Two related defects in the usage-limit retry path, both observed live — after a session limit reset, the monitor injected **5 spurious `Continue…` messages into an actively-working session**.

**1. No `isWorking` gate.** After a successful resume, the limit banner can linger in the captured scrollback (and a banner re-printed by another process keeps it visible the whole time), so the `'waiting'` path kept re-sending the retry message every poll up to `maxRetries` — driving a session that had already resumed. Fix: if the limit cleared **or** Claude is working again, stop and reset. The overload path already had this guard; this brings the usage path in line.

**2. Whole-capture matching.** `isRateLimited` matched the banner *anywhere* in the 20-line capture, so limit text merely **quoted in scrollback** (e.g. a conversation about rate limits, or a stale banner the session scrolled past) triggered a false wait/retry. `isRateLimited` now takes an optional `tailLines` arg; the monitor's *action* decisions only consider the live tail (last 12 lines), mirroring the overload path's tail-anchoring. Print mode keeps the full scan (captured process output, not a scrolling TUI).

+3 regression tests (191 total, zero deps).

### DIFF
--- a/src/monitor.js
+++ b/src/monitor.js
@@ -7,6 +7,10 @@ import { readStopFailureEvent, clearStopFailureEvent } from './events.js';
 
 const DEFAULT_FOREGROUND_COMMANDS = ['node', 'claude', 'npx', 'tsx', 'bun', 'deno'];
 const SHELL_COMMANDS = ['bash', 'zsh', 'sh', 'fish', 'dash', 'ksh'];
+// Only a usage-limit banner in the live tail counts — quoted limit text in scrollback
+// (a conversation about limits) or a banner the session already scrolled past is not the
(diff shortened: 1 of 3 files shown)


cheapestinference:
✅ Shipped in **v0.5.1** — `npm i -g claude-auto-retry@0.5.1`. The usage-retry anti-spam fix (tail-anchored detection + `isWorking` gate) is live. Thanks a lot @romerjon — this was a solid catch with great regression tests. 🙏 https://github.com/cheapestinference/claude-auto-retry/releases/tag/v0.5.1

PR #30, opened 2026-07-03
feat(tmux): add status bar indicator for monitor state

MedicD21:
## Summary

- The monitor's `monitoring`/`waiting`/`overload` state lives entirely in-memory inside a detached, unref'd process — there's no way to tell from outside whether a given pane is currently being watched, without tailing logs.
- This adds a per-pane status snapshot (`~/.claude-auto-retry/status/<pane>.json`), written on every poll tick, and a dependency-free `bin/tmux-status.sh` reader that renders it as a tmux status-bar segment:
  - actively monitoring → `🟢AR`
  - waiting on a usage-limit reset → `⏳AR 1h30m`
  - backing off from overload → `🟠AR 45s`
  - no monitor for the pane, or a stale file (monitor died without cleanup, >30s old) → nothing
- New script is registered as `claude-auto-retry-tmux-status` in `package.json` bin, so it lands on `PATH` after a normal `npm i -g`.
- Mirrors the existing pane-keyed write/read/clear shape from `events.js` (StopFailure markers) for consistency.
- Docs: new "tmux status bar indicator" README section + CHANGELOG entry under `[Unreleased]`.

## Test plan

- [x] `npm test` — all 201 tests pass (9 new in `test/status-file.test.js`, covering round-trip, sanitized filenames, absent/corrupt files, overwrite, and the no-op-without-a-pane-key guard)
- [x] Manually verified end-to-end against a real tmux pane + a live `startMonitor` process: status file appears, `tmux-status.sh` renders the correct icon for `monitoring`/`waiting`/`overload`, and both missing and stale (>30s) files render blank
- [x] Confirmed `git diff --cached --summary` preserves the executable bit on `bin/tmux-status.sh` (`100755`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   to the `API Error` render (mentioning the phrases in conversation can't
   trigger it), and the retry budget is kept across working ticks so a sticky
   flag stays bounded (#33).
+- tmux status bar indicator: the monitor now writes a per-pane status snapshot to
+  `~/.claude-auto-retry/status/<pane>.json` on every poll tick, and a new
+  `claude-auto-retry-tmux-status` script renders it as a status-bar segment
+  (`🟢AR` monitoring, `⏳AR 1h30m` waiting on a reset, `🟠AR 45s` overload backoff,
+  `🔴AR` gave up). Dependency-free POSIX shell; hides itself if a pane has no
+  monitor or its status file is stale (staleness is derived from the monitor's
+  actual poll interval, not a fixed constant).
 
 ### Fixed
 - `rate_limit` StopFailure events are no longer routed through the seconds-scale
--- a/README.md
+++ b/README.md
@@ -312,27 +312,60 @@ Usage limits always take precedence; the safeguard path only acts when Claude is
 (no `esc to interrupt` footer) and the foreground process is `claude`/`node`.
 ## tmux status bar indicator
 
-The monitor writes a small JSON snapshot to `~/.claude-auto-retry/status/<pane>.json`
-on every poll tick, so you can tell at a glance — without checking logs — whether a
-pane is being watched, waiting out a usage-limit reset, or backing off from overload.
+The monitor writes a small JSON snapshot per pane on every poll tick, so you can tell
+at a glance — without checking logs — whether a pane is being watched, waiting out a
+usage-limit reset, backing off from overload or a safeguard flag, or has given up.
 
 Add a segment to `status-right` (or `status-left`) in `~/.tmux.conf` that shells out to
-the bundled reader script, passing the current pane id:
+the bundled reader script, passing the current pane id **and** the server's socket path:
 
 ```tmux
 set -g status-interval 5
-set -g status-right "#(claude-auto-retry-tmux-status '#{pane_id}') | %Y-%m-%d %H:%M"
+set -g status-right "#(~/.local/lib/node_modules/claude-auto-retry/bin/tmux-status.sh '#{pane_id}' '#{socket_path}') | %Y-%m-%d %H:%M"
 ```
 
-`tmux` substitutes `#{pane_id}` with the *attached client's current pane* before running
-the command, so the segment always reflects whichever pane you're looking at. It prints:
+**Use an absolute path, not the bare command name.** `#()` commands run inside the tmux
+*server's* own environment, not the environment of whichever shell you attached from —
+if the server was started before your shell rc added `npm`/`nvm`'s bin directory to
+`PATH` (e.g. tmux auto-started at login, or by another program), the bare command name
+resolves to nothing and the segment stays permanently blank with no error anywhere.
+Find your actual install path with `which claude-auto-retry-tmux-status` (run it in a
+normal shell, then hardcode that path in `.tmux.conf`) if it differs from the example
+above. If you use nvm and switch Node versions, re-check the path.
+
+`tmux` substitutes `#{pane_id}` and `#{socket_path}` itself before running the command
+(these are tmux format variables, resolved at expansion time — not environment
+variables the script has to go looking for), so the segment always reflects whichever
+pane you're looking at, correctly scoped to the tmux server it belongs to. The
+`socket_path` argument matters if you ever run more than one tmux server on the same
+machine (e.g. `tmux -L work`, `tmux -L personal`, or two users' default servers on a
+shared host): pane ids like `%2` are only unique *within* a server, so without it two
+different servers' `%2` panes would render each other's status. It's technically
+optional (older single-argument configs still work, falling back to a shared key), but
+pass it unless you're certain you'll only ever run one tmux server.
+
+It prints:
 
 | Pane state | Indicator |
 |------------|-----------|
 | Actively monitoring | `🟢AR` |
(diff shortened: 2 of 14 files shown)


cheapestinference (changes requested):
I went through this carefully against the branch code. Useful feature, cleanly built (atomic tmp+rename writes, good module tests). The issues are all in the writer↔reader contract; most-severe first.

**1. Staleness threshold (30s) breaks with configurable poll intervals — CONFIRMED** (`bin/tmux-status.sh:25`)
`pollIntervalSeconds` has a min clamp but no max (`validNumber(val, 1, …)`), and the status file is written once per tick. With `pollIntervalSeconds: 60`, a healthy monitor's segment blanks out for ~half of every interval — precisely during an hours-long usage wait when it's most useful. Suggest writing the interval (or a TTL) into the JSON and deriving staleness from it, e.g. `age > 2×interval`.

**2. Give-up states render as live states — CONFIRMED** (`src/monitor.js`)
All three terminal give-up paths are indistinguishable in the snapshot: after `max-retries` the state stays `waiting` with `waitUntil` re-bumped every cycle (perpetually resetting ⏳ countdown); after the scraper overload cap it stays `overload` (🟠 rolling ~60s forever); the event-path cap leaves `monitoring` (green 🟢 for a monitor that will never retry again). A `gaveUp: true` field + a distinct indicator (e.g. `🔴AR`) would make the bar honest.

**3. No cleanup on signal/crash — CONFIRMED** (`src/monitor.js`, ~line 352)
`clearStatus` runs only on the `exit` result and the 10-consecutive-errors path; there is no `process.on('SIGTERM'/'SIGINT')` handler and no GC of the status dir. A killed monitor shows a false 🟢AR for up to 30s, and `<pane>.json` files from dead monitors accumulate forever (tmux pane ids grow monotonically per server). A signal handler plus a cheap best-effort sweep of stale files (on monitor start, or in `uninstall`) covers it.

**4. Pane-id collision across tmux servers — CONFIRMED, inherited design** (`src/status-file.js:16`)
Pane ids are unique per tmux *server*, not per machine: `tmux -L work` and `tmux -L personal` both have a `%2`, and both map to `_2.json` — one server's status bar renders the other server's monitor state. events.js has the same keying, but this PR makes it user-visible. Worth at least a README caveat; keying by `$TMUX` socket + pane would fix it.

**5. `#!/usr/bin/env bash` vs the "pure POSIX, dependency-free" claim — CONFIRMED** (`bin/tmux-status.sh:1`)
The body genuinely is POSIX (no bashisms), so the bash dependency is gratuitous and breaks silently on bash-less hosts (Alpine/busybox — platforms the installer explicitly supports). `#!/bin/sh` makes the README claim true.

**6. README snippet may silently resolve nothing in tmux's `#()` environment — PLAUSIBLE** (`README.md`, tmux config section)
`#()` runs with the tmux *server's* environment; nvm/npm-prefix bin dirs are added by shell rc, so a server started outside such a shell has no `claude-auto-retry-tmux-status` on PATH → permanently blank segment, no error anywhere. The script's own header comment already shows absolute-path usage; the README should too (or mention PATH).

Cleanup (both CONFIRMED, non-blocking): `src/status-file.js` is a near-verbatim copy of `events.js`'s pane-keyed file mechanism (the sanitizer rule now lives in three places, including the `tr` call in the shell reader) — worth extracting a shared helper; and `writeStatus` runs `mkdir(recursive)` on every 5s tick — hoist it.

cheapestinference:
Heads-up: #31 and #33 just merged and this branch now conflicts in `src/monitor.js` (the startMonitor loop / result-logging area both PRs touched). Please rebase onto master when addressing the review — the new `safeguard` status is also worth including in the snapshot's status values while you're in there.

MedicD21:
Rebased onto `master` (the `src/monitor.js` conflict from #31/#33 was just the import lines — kept both) and addressed all six points, most-severe first:

1. **Staleness threshold** — every snapshot now carries `pollIntervalSeconds`; the reader derives staleness as `2× interval` instead of a fixed 30s.
2. **Give-up states rendering as live** — added a `gaveUp` flag, set at all four terminal give-up paths (usage max-retries, overload scraper cap, overload event-path cap, safeguard cap) and cleared on recovery. Renders as a distinct `🔴AR`, overriding the stale `🟢`/`⏳`/`🟠`.
3. **No cleanup on signal/crash** — added `SIGTERM`/`SIGINT` handlers that clear the status file, plus a best-effort `sweepStaleStatus()` GC on monitor start and on `uninstall` (SIGKILL still can't be caught, as noted).
4. **Pane-id collision across servers** — status files are now keyed by tmux socket path + pane id. Reader takes `#{socket_path}` as a second arg (README updated); old single-arg configs still work via a fallback key.
5. **Shebang** — changed to `#!/bin/sh`.
6. **README PATH caveat** — added, with an absolute-path example and `which claude-auto-retry-tmux-status` to find the real path.

Also did the suggested cleanup (shared `src/pane-key.js` sanitizer used by both `events.js` and `status-file.js`; hoisted the `mkdir`), and added the new `safeguard` status to the snapshot and the reader's case statement (`🛡AR`).

All 237 tests pass, including new coverage for the socket-scoping, the sweep, and each `gaveUp` path.

cheapestinference:
Thanks for the rework since the conflict heads-up — the socket-scoping, signal cleanup, interval-aware staleness, and honest give-up indicator are all good additions. The tricky shell bits are correct: the `[ -z "$status" ] || [ -z "$updated" ] && exit 0` precedence does the intended `(A || B) && C` under POSIX, and the arithmetic doesn't crash on missing/empty fields. Two minor bugs:

**1. The single-argument (socket-less) config never renders.** The monitor is `fork()`ed inside the pane and always inherits `$TMUX`, so `writeStatus` always name-prefixes the file with the real socket path (e.g. `…tmux-1000_default__2.json`). A single-arg reader falls back to `socket="default"` and looks for `default__2.json`, which never matches → the segment stays blank forever. The README ("older single-argument configs still work, falling back to a shared key") and the script header comment both describe a shared-key fallback that doesn't exist, since the writer never uses `default`. Either make the writer honor a shared key when `$TMUX` is unset, or drop the backward-compat claim and document the 2-arg form as required.

**2. `sweepStaleStatus` never removes orphaned `.tmp` files.** The sweep loop guards with `if (!name.endsWith('.json')) continue;`, but `writeStatus` names its temp file `<key>.json.<pid>.tmp` — which ends in `.tmp`, so it's skipped. The `catch` block's comment ("an orphaned .tmp file from a crash mid-rename — remove it too") is unreachable: a `.tmp` file never passes the `.json` filter to reach it. Net effect is that `.tmp` files from a crash mid-rename accumulate forever, contradicting the "directory doesn't grow unbounded" rationale. A separate `.tmp` + age check, or matching both suffixes, fixes it.

Minor observation (not necessarily for this PR): when the overload path hits the cumulative-wait cap it returns the give-up state but leaves `status='monitoring'` and never clears `overloadTotalWaitMs`/`_gaveUp` (only `resetOverload` does, which that path doesn't reach). So if Claude later self-recovers and keeps working, the segment can stay `🔴AR` even though the pane is healthy. Looks like a pre-existing state-machine quirk the new flag just surfaces.

cheapestinference:
Merged — thanks @MedicD21. Both findings were addressed cleanly (the `.tmp` orphan sweep with an age guard, and the honest `#{socket_path}`-required docs), and the full suite is green (239/239) on merge with master. Nice, self-contained feature.

PR #32, opened 2026-07-03
feat(cli): add `reconcile` — re-arm monitors for all live claude panes

romerjon:
**Problem.** Monitors are detached processes with no supervising service. If one is killed/crashes — or a claude session is started outside the wrapper — that live session is silently left unmonitored. Only *new* wrapper-launched sessions get a monitor, so coverage degrades over time with no way to restore it short of manual `pgrep`/`kill`/relaunch (error-prone: pane IDs get reused).

**`reconcile`.** Re-arms a monitor for every live tmux pane running claude that isn't already covered, from authoritative tmux + process state:
- maps each live claude to its tmux pane by walking the process tree to the pane's `pane_pid`;
- skips panes that already have a monitor (idempotent — safe to run anytime / on a timer);
- never arms its own pane (`$TMUX_PANE`);
- handles **pane-id reuse** (two claudes resolving to one recycled pane → prefer the foreground one, else highest pid);
- `--dry-run` previews the plan.

The pure parse/plan core is separated from the tmux/ps/spawn runner and unit-tested (11 tests). Uses `spawn` (not `fork`) for the detached monitors so no IPC channel keeps the CLI alive — it exits immediately after arming.

Usage:
```
claude-auto-retry reconcile            # re-arm all uncovered live sessions
claude-auto-retry reconcile --dry-run  # preview
```

Pairs well with a periodic `--user` timer to make monitor coverage self-healing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,19 @@ All notable changes to this project are documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.5.1] - 2026-06-30
+
+**Upgrade if you installed `0.5.0` from npm.** The `0.5.0` npm artifact was built
+before #29 was merged and shipped without the usage-retry anti-spam fix. `0.5.1`
+includes it. (The git tag `v0.5.0` already contained #29; only the npm tarball was
+behind.)
+
+### Fixed
+- Stop the usage-retry path from spamming an already-resumed session: a lingering
+  limit banner in scrollback no longer re-injects `Continue…` every poll. Detection
+  is now anchored to the live tail, and an `isWorking` gate stops the moment Claude
+  resumes (#29).
+
 ## [0.5.0] - 2026-06-30
 
 This release rolls up everything merged since `0.2.2`, including the API
@@ -18,10 +31,6 @@ overload backoff engine and interactive `/rate-limit-options` menu navigation.
 - Enable mouse scroll and vi copy-mode in tmux sessions created by the tool (#25).
 
 ### Fixed
-- Stop the usage-retry path from spamming an already-resumed session: a lingering
-  limit banner in scrollback no longer re-injects `Continue…` every poll. Detection
-  is now anchored to the live tail, and an `isWorking` gate stops the moment Claude
-  resumes (#29).
 - Require Claude to be in the foreground before driving the
   `/rate-limit-options` menu, preventing keystrokes from leaking into the wrong
   pane (#28).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   to the `API Error` render (mentioning the phrases in conversation can't
   trigger it), and the retry budget is kept across working ticks so a sticky
   flag stays bounded (#33).
+- tmux status bar indicator: the monitor now writes a per-pane status snapshot to
+  `~/.claude-auto-retry/status/<pane>.json` on every poll tick, and a new
+  `claude-auto-retry-tmux-status` script renders it as a status-bar segment
+  (`🟢AR` monitoring, `⏳AR 1h30m` waiting on a reset, `🟠AR 45s` overload backoff,
+  `🔴AR` gave up). Dependency-free POSIX shell; hides itself if a pane has no
+  monitor or its status file is stale (staleness is derived from the monitor's
+  actual poll interval, not a fixed constant).
 
 ### Fixed
 - `rate_limit` StopFailure events are no longer routed through the seconds-scale
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,35 @@
+# Changelog
+
+All notable changes to this project are documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [0.5.0] - 2026-06-30
+
+This release rolls up everything merged since `0.2.2`, including the API
+overload backoff engine and interactive `/rate-limit-options` menu navigation.
+
+### Added
+- Detect sustained API overload (`529`/`500`/`503`) and retry with exponential
+  backoff, including an event-driven (`StopFailure`) mode (#20, hardened).
+- Interactive navigation of the `/rate-limit-options` menu, driving it to
+  "Stop and wait" across any menu layout (#19, #26).
+- Enable mouse scroll and vi copy-mode in tmux sessions created by the tool (#25).
+
+### Fixed
+- Stop the usage-retry path from spamming an already-resumed session: a lingering
+  limit banner in scrollback no longer re-injects `Continue…` every poll. Detection
+  is now anchored to the live tail, and an `isWorking` gate stops the moment Claude
+  resumes (#29).
+- Require Claude to be in the foreground before driving the
+  `/rate-limit-options` menu, preventing keystrokes from leaking into the wrong
+  pane (#28).
+- Reliable retry submission plus session/weekly rate-limit detection (#7, #15, #22).
+- Correct an off-by-a-day wait when parsing reset times in offset timezones (#6, #23).
+- Unalias `claude` before defining the wrapper, fixing a zsh/bash `source` error (#10, #24).
+- Skip send-keys correctly when the foreground process is the shell, not Claude (#1).
+
+## [0.2.2] - 2026-03-31
+
+- Last published baseline release.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,23 @@ All notable changes to this project are documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+- Safeguard/AUP false-positive auto-retry: when the model's safeguards flag a
+  message ("safeguards flagged this message"), re-send a short retry up to
+  `safeguard.maxRetries` times, then give up loudly once. Detection is anchored
+  to the `API Error` render (mentioning the phrases in conversation can't
+  trigger it), and the retry budget is kept across working ticks so a sticky
+  flag stays bounded (#33).
+
+### Fixed
+- `rate_limit` StopFailure events are no longer routed through the seconds-scale
+  overload path — a session/usage limit is an hours-scale wait owned by the
+  usage path, and the misroute made the two fight (futile `Continue` retries
+  into a session-limited pane). The marker error type is validated at the
+  consumer too, so an outdated installed hook can't reintroduce it (#31).
+
 ## [0.5.1] - 2026-06-30
 
 **Upgrade if you installed `0.5.0` from npm.** The `0.5.0` npm artifact was built
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- **`reconcile` / `install-timer` / `exclude-self`** for self-healing monitor coverage: a
+  monitor killed (or a `claude` started outside the wrapper) is re-armed from live tmux +
+  process state, on demand or via a `systemd --user` timer (#32).
 - Safeguard/AUP false-positive auto-retry: when the model's safeguards flag a
   message ("safeguards flagged this message"), re-send a short retry up to
   `safeguard.maxRetries` times, then give up loudly once. Detection is anchored
@@ -23,11 +26,29 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   actual poll interval, not a fixed constant).
 
 ### Fixed
+- `install-timer` no longer crashes on npm installs — `systemd/` is shipped in the package
+  and the template reads fail with a clear message instead of ENOENT (#32).
+- `reconcile` distinguishes a real `pgrep` failure (ENOENT, busybox without `-a`, macOS
+  PID-only output) from "no monitors running", and aborts loudly rather than arming a
+  duplicate monitor per pane every run (#32).
+- Monitor coverage is keyed per-pane, so a stopped `claude` keeping its monitor can't lead
+  to a second monitor on the same pane (#32).
+- A single-instance lock (pid + start-token identity) stops an overlapping manual + timer
+  run from double-spawning, and can't wedge on PID reuse (#32).
+- Exclude-file PID entries are pruned when dead, so kernel PID reuse can't permanently mute
+  a future session (the self-expiring behavior the docs promised) (#32).
+- Print-mode panes (`claude -p` / `--print`) are no longer given a send-keys monitor (#32).
+- The generated systemd unit quotes the node/CLI paths (spaces no longer break it), drops
+  the no-op `Persistent=true`, and `install-timer` prints an nvm re-run caveat (#32).
 - `rate_limit` StopFailure events are no longer routed through the seconds-scale
   overload path — a session/usage limit is an hours-scale wait owned by the
   usage path, and the misroute made the two fight (futile `Continue` retries
   into a session-limited pane). The marker error type is validated at the
   consumer too, so an outdated installed hook can't reintroduce it (#31).
+
+### Changed
+- Removed the dead `CLAUDE_COMMANDS` constant; documented that reconcile matches
+  `comm === 'claude'` (a bare-`node` session without `process.title` isn't detected) (#32).
 
 ## [0.5.1] - 2026-06-30
 
--- a/DESIGN-NOTES.md
+++ b/DESIGN-NOTES.md
@@ -1,0 +1,201 @@
+# Design notes
+
+Forward-looking design notes for `claude-auto-retry`, grounded in a research pass
+(2026-06) against the installed Claude Code binary (v2.1.195, decompiled), the
+Anthropic API error schema, and Claude Code's hooks/transcript surfaces. Ordered by
+leverage. Items marked **[done]** ship in the current version; the rest are proposals
+with enough detail to execute.
+
+## 0. The core problem: we scrape a human-facing render
+
+The overload path detects a *terminal* API error by scraping `tmux capture-pane` and
+string-matching. That is fundamentally a heuristic over output meant for humans, and
+every false positive we've hit (a `res.status(503)` under edit, an HTTP status code in
+tool output in scrollback) and every false negative we worry about is a symptom of that
+layer choice.
+The hardening below makes the scraper as good as a scraper can be — but the strategic
+move is to stop using the scrape as the *trigger*.
+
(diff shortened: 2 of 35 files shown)


cheapestinference (changes requested):
I reviewed this against the branch code, including empirical checks of `comm` behavior and the systemd semantics. The gap this closes is real, but as it stands the headline path is broken for npm installs and there are several verified routes to duplicate monitors — the exact double-injection class #29 fixed. Most-severe first.

**1. `install-timer` crashes on npm installs — CONFIRMED** (`package.json` / `bin/cli.js`)
The `"files"` whitelist (`bin/`, `src/`, LICENSE, README, CHANGELOG) doesn't include the new `systemd/` directory, and `cmdInstallTimer` `readFile`s the unit templates with no try/catch. `npm i -g claude-auto-retry && claude-auto-retry install-timer` → unhandled ENOENT. Works only from a git checkout. Add `"systemd/"` to `files` (or embed the templates in the JS).

**2. `gather()` treats any pgrep failure as "zero monitors running" — CONFIRMED** (`src/reconcile.js:146`)
The catch is unconditional (`catch { /* none running → pgrep exits 1 */ }`), so pgrep ENOENT, or busybox pgrep (no `-a`; Alpine is an explicitly supported platform) makes `running` an empty Map → the 5-minute timer arms one *additional* monitor per covered pane on every run, unbounded, each of them injecting on the next rate-limit. Bonus: on macOS/BSD `pgrep -af` succeeds but prints PIDs only, so `parseRunningMonitors` matches nothing and duplication happens silently with no error at all. Distinguish exit-code-1 from real failures, and verify the output shape.

**3. Coverage keyed on `pane pid` lets two persistent monitors share a pane — CONFIRMED** (`src/reconcile.js:129`)
A Ctrl-Z'd (SIGSTOP'd) claude keeps its monitor alive (`kill(pid, 0)` succeeds on stopped processes), and reconcile then arms a second monitor for the new foreground claude in the same pane. Both pass the foreground gate (pane_current_command = the new claude) and both send retry keys on a banner. Keying coverage per-pane (one monitor per pane, whatever the pid) avoids the whole class.

**4. No lock / single-instance guard — CONFIRMED** (`src/reconcile.js`)
`gather → plan → arm` samples coverage once before spawning. A manual `reconcile` overlapping a timer fire both plan the same arm set and both spawn; nothing ever reaps extras (later runs just say "already monitored"). A simple lockfile (or arming idempotently by checking again right before spawn) fixes it.

**5. Exclude entries never expire, despite the code promising they do — CONFIRMED** (`src/reconcile.js:122,205`)
`excludeSelf` appends PIDs that are never pruned, and the match is a bare string compare with no liveness/start-time validation. The header comment ("SELF-EXPIRING — … can never mute a different future session"), the README, and the command's own output all state the opposite. Kernel PID reuse (pid_max 32768 on many distros) will eventually hand a stale excluded PID to a brand-new claude, which the timer then skips forever. Prune dead PIDs on read, or store pid+starttime.

**6. `CLAUDE_COMMANDS = ['claude', 'node']` is dead code masking a coverage gap — CONFIRMED** (`src/reconcile.js:46,103`)
Nothing references the constant; matching is `p.comm === 'claude'` only. Verified empirically: a `#!/usr/bin/env node` shebang script that doesn't set `process.title` shows comm `node` and is invisible to reconcile — silent zero coverage (`No tmux panes running claude found`) for exactly the installs this feature claims to protect. Either honor the constant or delete it and document the limitation.

**7. Generated systemd unit is fragile — CONFIRMED** (`bin/cli.js:279` / `systemd/*.timer`)
`ExecStart=__NODE_PATH__ __CLI_PATH__ reconcile` is substituted unquoted (any path with a space breaks deterministically) and bakes the absolute `process.execPath` — after an nvm upgrade the timer fails on every fire, silently (no warning is printed, unlike `cmdInstall`'s nvm note). Also `Persistent=true` is a no-op without `OnCalendar=` (monotonic timers ignore it), so the "catch up on a missed run" comment describes behavior that never happens.

**8. `claude -p` (print mode) panes get a send-keys monitor — PLAUSIBLE** (`src/reconcile.js:103`)
The wrapper deliberately never arms a monitor in `launchPrintMode`; reconcile can't tell (comm carries no args) and will arm one, risking retry text injected into piped/scripted output. Filtering out processes whose argv contains `-p`/`--print` (via `ps -o args`) would cover it.

Non-blocking cleanup the finders also flagged: `MONITOR_PATH` and the current-pane lookup duplicate `launcher.js`/`tmux.js` helpers; `cmdExcludeSelf` re-runs pgrep with its own parser instead of reusing `parseRunningMonitors`; a couple of `await (await import('node:fs/promises'))` where the module is already statically imported.

romerjon:
Thanks — all eight reworked test-first (failing test driving the real code, then the fix). Suite green (227 tests). Commits are per-finding.

**1 — install-timer crashes on npm installs.** Added `"systemd/"` to package.json `files`, and `cmdInstallTimer` now reads the templates inside a try/catch that prints a clear "upgrade / run from a checkout" message instead of an unhandled ENOENT. (`f765e5f`)

**2 — pgrep failure vs "zero monitors".** Extracted pure `runningFromPgrep(err, stdout)`: only exit-code-1-with-no-output means "nothing running"; ENOENT, any non-1 exit (busybox usage error), or a success whose output has no parseable monitor args (macOS/BSD PID-only) throws, so `gather()` aborts loudly instead of arming a duplicate per pane. (`46340fd`)

**3 — per-pane coverage.** Coverage is keyed per-pane now (`coveredPanes` from the running-monitor keys), so a SIGSTOP'd claude keeping its monitor can't cause a second monitor on the same pane. (`71d1d9f`)

**4 — no lock.** Added an atomic `O_CREAT|O_EXCL` lockfile (`acquireLock`); a real run holds it, a concurrent run returns `{ locked: true }` and the CLI reports it. Crash-safe: a stale lock is stolen only when its holder pid is dead, so it can never wedge the timer. (`ef04b06`)

**5 — exclude entries never expire.** `readExcludeFile` now prunes dead numeric PIDs via pure `pruneExcludeEntries` (EPERM treated as alive), keeping pane ids and live pids — so kernel PID reuse can't mute a future claude. Fixed the README/header claims to describe the real mechanism. (`0ff11f5`)

**6 — dead CLAUDE_COMMANDS.** Removed; replaced with a documented `CLAUDE_COMM` + a comment stating the limitation (a session that doesn't set `process.title` to `claude` shows comm `node` and isn't detected — matching bare `node` would arm on every node process). (`dbc3e78`)

**7 — fragile systemd unit.** ExecStart placeholders are quoted (`\"__NODE_PATH__\" \"__CLI_PATH__\"`, substitution extracted to `renderReconcileUnit`); dropped `Persistent=true` (no-op on a monotonic timer) with a corrected comment; `install-timer` prints an nvm re-run caveat like `install` does. (`f765e5f`)

**8 — print-mode panes.** ps now captures args (`-eo …,args=`), `parseProcesses` exposes them, and both `planReconcile` and `claudePidForPane` filter out `-p`/`--print` sessions (anchored so `-pretty` doesn't false-match). (`dbc3e78`)

**Cleanup** (`ef04b06`, `9b32b83`): folded `excludeSelf`'s dynamic `fs/promises` imports into the static import; `cmdExcludeSelf` reuses `parseRunningMonitors` instead of a bespoke parser.

Smoke-tested `reconcile --dry-run` against a live box (8 panes): per-pane coverage, self-exclusion, and the new args parsing all behave.

cheapestinference:
Really nice work — the pure/impure split makes this reviewable, and the headline risks all hold up: no duplicate-per-fire (wrapper-armed monitors are correctly seen as covered, and any pgrep ambiguity throws rather than reporting a false "zero"), per-pane coverage keying, `PRINT_MODE` anchoring, pid-vs-pane exclusion, and the pane-id-reuse foreground preference all check out.

Two issues left, both in `acquireLock` (`src/reconcile.js:50-64`), same root cause: the lock file stores only a bare PID (`fh.writeFile(String(process.pid))`), with no boot-id/start-time, so a *stale* lock (a reconcile SIGKILLed mid-run before the `finally` release — OOM killer, etc.) can't be distinguished from a live holder by PID alone.

**1. PID-reuse wedge → self-healing silently disabled.** If a stale lock holds PID `12345` and the kernel later reuses `12345` for any unrelated live process, `isProcessAlive(12345)` returns true on every call (it also treats `EPERM` as alive), so `acquireLock` returns `{ ok:false }` forever. `reconcile()` returns `{ locked:true }` and the timer becomes a permanent no-op — monitors are never re-armed again, with no error surfaced.

**2. Stale-lock steal is a TOCTOU.** Two concurrent runs (the exact overlapping manual+timer case the lock defends against) can both read `holder=dead`, both `unlink`, and both then `open('wx')` successfully → both hold the lock → the double-spawn the lock exists to prevent. Related: `release()` unlinks by path unconditionally, so once two runs both "hold" it, one's release deletes the other's lock.

Both are low-probability (need a rare stale lock), so not blockers, but worth hardening. Cheapest fix is to make the holder identity unforgeable across PID reuse — write `pid + start-time` (from `/proc/<pid>/stat` field 22, or boot-id) and compare both before deciding a lock is stale — or drop the manual steal path and rely on `flock`/`O_EXCL` with a TTL. Happy to take a follow-up PR for this if you'd rather land reconcile first.

romerjon:
Both `acquireLock` issues addressed, plus rebased onto current master.

**Lock identity is now pid + start-token** (`<pid>\t<startToken>`), where the start token is the process start time — `/proc/<pid>/stat` field 22 on Linux, `ps -o lstart=` fallback elsewhere:

- **PID-reuse wedge** — `holderIsLive` treats a live PID whose start token no longer matches as stale (the PID was reused), so a stale lock can never wedge acquisition at `{ok:false}` forever. A dead PID is stale; a legacy bare-PID lock is respected-if-present (can't disambiguate, so don't risk stealing a real holder).
- **Steal TOCTOU + cross-delete** — `release()` now only unlinks when the file still holds *our* identity, so a run that was stolen from can't delete the new holder's lock. And acquire re-reads after `O_EXCL` create to detect a concurrent clobber and back off rather than double-hold.

Tests cover: PID-reuse steal, a genuine live holder with a matching token being respected (not stolen), and no cross-delete on release.

**Rebased onto current master** — resolved against the #30 tmux-status/status-file work (CHANGELOG/README/`package.json`/`cli.js`). Full suite **277 green**, `reconcile --dry-run` smoke-tested against a live box.

Note: I kept the residual "two concurrent runs both steal a stale lock in the exact same microsecond window" as a very narrow edge (needs a crash-orphaned lock *and* precise interleaving) — the re-read-after-create shrinks it but doesn't fully close it without a rename-CAS. Happy to go further (rename-based CAS, or `flock`) if you'd like, but it seemed past the point of diminishing returns for a 5-minute timer.

cheapestinference:
Verified `7c05dcd3` against the real exported `acquireLock`. The pid+start-token identity is solid — finding 1 is genuinely fixed — but finding 2 (the steal TOCTOU) is narrowed, not closed: I reproduced a double-hold against this branch. One item to fix, two minor notes.

### What holds

- **`/proc/<pid>/stat` parsing is correct**, including the classic trap: you slice past the *last* `)` before splitting, so a comm containing spaces and parens parses fine. I verified empirically with a process named `(ev il) ( x)` — the token came back as exactly field 22, stable across reads, and the same function is used at write and check time.
- **PID-reuse wedge is fixed and well-tested**: live PID + mismatched token → stolen; live PID + matching token → respected; dead bare-PID legacy lock → stolen and rewritten in the new format (the original wedge does not recur). The only residual legacy case — a dead bare-PID holder whose PID was reused by a live process — is undecidable without a token, one-time, self-resolving, and documented. Fine.
- Release-side no-cross-delete works. Suite green (277/277), rebase onto current master checks out.

### Still open: acquire-side steal can delete a *live* lock (double-hold reproduced)

The no-cross-delete guard exists only in `releaseLock`. The steal path in `acquireLock` is an unconditional `unlink` of whatever is at the path — including a lock a racer just legitimately created:

1. Stale lock S sits at the path. A and B both hit EEXIST and both read S (stale).
2. A unlinks, creates, writes, re-reads its own content → `ok: true`.
3. B, slightly delayed, unlinks **A's fresh live lock**, creates, writes, re-reads → `ok: true`. Both hold.

The re-read only catches clobbers that land *before* the re-read. There's a second window with no stale lock at all: `open('wx')` + `writeFile` aren't atomic, so a contender reading between create and write sees an empty file, treats the falsy holder as stale, and unlinks the winner's just-created lock — that's the plain timer+manual-run overlap the lock exists for.

Repro (8 concurrent processes calling the real `acquireLock`, winners holding for 1.5s so a simultaneous hold is unambiguous): 1 double-hold in 120 iterations seeded with a stale lock, and 1 in 120 with *no* pre-existing lock. Rare, but the consequence is exactly the original finding — duplicate monitors armed per pane, both sending keys.

Cheap standard fix that closes both windows:

- **Create atomically with content**: write `pid\ttoken` to a unique temp file, then `link(tmp, lockPath)` (or `rename` with `O_EXCL` semantics via link) — there is never an empty lock visible.
- **Steal via `rename(lockPath, uniqueGraveyardName)`** instead of unlink: rename of the same source path succeeds for exactly one stealer; the loser's rename ENOENTs and it retries. Re-read the *renamed* file to confirm you stole the stale lock you evaluated (identity check), then create. No unconditional deletes anywhere.

(Also observed: 2/120 stale-seeded runs ended with zero winners — everyone backed off after 2 attempts, leaving a dead-PID lock the next run steals. One-cycle liveness blip, self-healing; the rename approach with a retry loop happens to fix this too.)

### Two minor notes (non-blocking)

- **Null-token asymmetry**: if both `/proc` and `ps -o lstart=` fail (busybox-without-procfs class), `myId` is `"<pid>\t"` but the re-read `.trim()`s the trailing tab, so the verify never matches → `acquireLock` permanently returns `{ok:false}` *and* leaves its lock file behind. One-line fix: compare trimmed forms.
- **`ps lstart` is locale-dependent** (LC_TIME), so on non-Linux a checker under a different locale env (cron vs shell) than the writer would misjudge a live holder stale. Linux always takes the `/proc` path on both sides so it's unaffected, and this tool targets Linux/systemd — just worth a comment if you keep the fallback.

With the rename-based steal in place this is mergeable from my side — everything else from both rounds now holds.

romerjon:
Closed the steal race with the standard link/rename approach you suggested.

- **Create atomically with content**: write `pid\ttoken` to a private temp file, then `link(tmp, lockPath)` — fails EEXIST if a lock exists, and no racer ever sees the empty/half-written window that `open('wx')`+write left.
- **Steal via `rename(lockPath, graveyard)`**: rename of a path succeeds for exactly one racer; the loser ENOENTs and retries. After winning, I re-read the graveyard file and, if a racer replaced the lock with a *live* one in the gap, restore it rather than steal it — no unconditional unlink anywhere.
- **Null-token asymmetry fixed**: the id is now a bare `<pid>` when no token is available (not `<pid>\t`), so `release`'s trimmed compare matches its own content — no more leak-and-wedge.
- **`ps lstart` locale caveat** documented (Linux takes the /proc path both sides, so unaffected).

Added concurrency regressions: 8 in-process acquirers → exactly one winner, both seeded with a stale lock and unseeded; stable across repeated runs. (In-process contention can't fully mimic your 8-process repro, but it exercises the interleaving and the primitives are atomic.) Suite green (279).

cheapestinference (changes requested):
Verified `bc6d4d1` against the real `acquireLock`. The null-token id fix and the `link()`-based atomic create are solid, and the 2-way stale steal now grants exactly one winner — nice. But the double-hold this review was about still reproduces, now through the **restore** branch.

The window: after `rename(lockPath → grave)` succeeds, `lockPath` is absent while we `readFile(grave)` + `holderIsLive(stolen)` — both `await`s, so the event loop yields there. A concurrent `link(tmp, lockPath)` creator can legitimately repopulate `lockPath` in that gap, and the restore then does `rename(grave, lockPath)`, which *atomically replaces* the destination — clobbering that fresh holder while it still believes it owns the lock.

Interleaving (A = stale/dead holder, B = recreator, C = stealer, D = fresh victim):
1. `lockPath` = A (stale). C reads holder → A, decides to steal.
2. Before C acts, B steals A and `link()`s a fresh **live** lock → `lockPath` = B.
3. C does `rename(lockPath → grave_C)` — moves **B's live lock** into `grave_C`. `lockPath` now empty.
4. In the async gap (C's `readFile(grave_C)` + `holderIsLive(B)`), **D** arrives, `link(tmp, lockPath)` succeeds → `lockPath` = D, D returns `{ok:true}`.
5. C reads `grave_C` = B; `B !== holderId(A)` && `holderIsLive(B)` → restore: `rename(grave_C → lockPath)` overwrites D's lock with B.
6. B holds (`{ok:true}` from step 2) **and** D holds (`{ok:true}` from step 4) — both in the critical section. Double-hold.

Reproduced with real processes / distinct PIDs and a mutual-exclusion detector: ~0.3–0.5% double-hold under contention, with restore-clobber events tracking double-holds roughly 1:1 (log lines show the restore overwriting a live, different-PID holder).

Fix direction: make the restore **non-clobbering** — e.g. `link(grave, lockPath)` + verify (fails `EEXIST` if a racer repopulated the path), or re-validate ownership after restoring, rather than an unconditional `rename` onto a path another run may have legitimately refilled.

Nit: the new regression tests call `acquireLock` 8× **in the same process**, so they share one `process.pid`/identity and never exercise cross-process syscall interleaving — losers just see their own live PID and back off. They stay green even with the bug above; a real repro needs child processes.

romerjon:
The non-clobbering restore wasn't enough — so I built your repro and measured, then changed approach.

**Measurement.** I wrote your cross-process harness (8 procs, seeded stale lock each round, `process.hrtime.bigint()` hold-interval overlap detector). With the non-clobbering `link()` restore in place, still **~10% time-overlapping holds** under seeded contention. Root cause is deeper than the clobber: once run C moves run B's *live* lock to the graveyard, B is a phantom holder (`ok:true`, no lock file) while a fresh creator D takes the emptied path — and no restore variant fixes that, because B already returned. Any steal that re-verifies liveness only *after* the destructive move is racy.

**New approach — serialize the break.** Creating the lock is a plain atomic `linkCreate` (temp + `link()`). A *stale* lock is removed and recreated only while holding a second **breaker** lock, so exactly one run ever breaks a given stale lock; everyone else backs off (live holder) or waits. No live holder's lock is ever moved, so there's no phantom. The breaker is itself `link()`-based and only cleared when its own holder is dead (crash-safe).

**Result: 0 time-overlapping holds across 600+ seeded-contention iterations** (5×120), stable. And the regression is now a **real cross-process test** — spawns processes, seeds a stale lock each round, asserts no overlapping hold intervals — since, as you noted, the in-process version couldn't exercise this. Suite green (278).

romerjon:
Two follow-up-review fixes (`ed45fdf`): **(F2, double-hold)** the breaker-cleanup `unlink` wasn't identity-checked — when the breaker holder was stale, it could delete a fresh LIVE breaker a racer created in the gap, breaking serialization → two runs both break the lock. Cleanup is now `releaseLock(breakerPath, bId)` (removes only if it still holds the dead id we read) + we re-confirm we still own the breaker before touching the main lock. Stress with BOTH main and breaker seeded stale (8 procs): 0 overlapping holds / 400 iters. **(F5)** `isPrintMode` now only counts `-p` as a flag before the first positional, so `claude "explain the -p flag"` isn't misread as print mode and left silently unmonitored.

cheapestinference:
Superseded by #42. I rewrote `master`'s history to clean up commit metadata, which changed the base commits this branch was built on and invalidated the diff here. Recreated as #42 — rebased onto the current `master`, **your commits and authorship preserved**, tests green, review carried over. Nothing needed from you; continue any follow-up on #42. Thanks again!

PR #33, opened 2026-07-03
feat(safeguard): bounded auto-retry past AUP-safeguard false positives

romerjon:
Adds a third recovery path alongside usage-limit waits and 5xx overload backoff: the model's **safeguards flagging a message**, which Claude Code renders as

```
● API Error: <model>'s safeguards flagged this message (…/legal/aup). They may flag
  safe, normal content as well. … Claude Code can't respond to this request with <model>.
  Double press esc to edit your last message, or try a different model with /model.
```

The error itself notes these "may flag safe, normal content" — they're often false positives and semi-random, so an **immediate re-send frequently clears them**. But a *sticky* flag (genuinely blocked content/model) must not loop.

**Behavior.** On detecting the render at an idle prompt, the monitor sends a short retry message (`continue`), waits `retryDelaySeconds`, and repeats — capped at `maxRetries` (default **3**), then gives up loudly (logged) rather than looping. Foreground-gated and tail-anchored (last 12 pane lines) like the overload path, so the phrase in scrollback or in a conversation *about* safeguards won't trigger it. Usage limits take precedence; a flag co-present while Claude is working is ignored.

Lifecycle (verified): `safeguard-detected → safeguard-retried ×3 → safeguard-gave-up` (exactly 3 sends).

**Config** — a new `safeguard` block with validation/fallbacks:
```json
{ "safeguard": { "enabled": true,
    "patterns": ["safeguards flagged this message", "can't respond to this request with", "legal/aup"],
    "maxRetries": 3, "retryDelaySeconds": 8, "retryMessage": "continue" } }
```

+20 tests; README "Safeguard retry" section. Zero deps.

PR #34, opened 2026-07-04
fix(patterns): chrome-aware tail so a limit banner behind a task widget still detects

romerjon:
**Symptom (observed live).** A genuine session-limit banner — `You've hit your session limit · resets 2am (Europe/Zurich)` + `/usage-credits to finish…` — sat ~16 lines above the prompt because a tall **todo/task widget**, the `✻ Brewed for 54m` status spinner, the input box, and the footer rendered below it. The tail-scoped `isRateLimited` (last 12 lines, added to fix a scrollback false-positive) never reached the banner, so **no retry fired and the session stalled ~54 min.**

**Why not just widen the window.** That reopens the false positive it was added for: a limit banner buried 15+ lines up *in scrollback* would re-match. Window size alone can't tell the two apart.

**Fix — measure distance in *content*, not raw lines.** The discriminator is what sits *below* the banner: after a live limit only Claude Code UI chrome renders below it (input box, footer, key hints, todo/task widget, status spinner, `/usage-credits` hint, background-agent notices); a stale/quoted banner has real output below it. So the detector strips **trailing chrome** before applying the tail window. A quoted banner with real work below stays out of the window (false-positive still fixed); a live banner behind a tall widget is now reached.

Plus a `/usage-credits` companion backstop scanned over a wider window — a distinctive live-limit string that catches the modern render even if a future widget escapes the chrome allowlist.

Details:
- `isWorking` deliberately keeps a **raw** tail (the `esc to interrupt` / `Retrying` indicators would otherwise be stripped as chrome).
- Same chrome-aware tail applied to the overload detector (same failure class).
- `·`/`•` excluded from the task-item chrome pattern (they also separate `· resets 3pm` in the multi-line render).
- Capture bumped 20→50 for headroom.

+10 tests including the exact reported pane; the scrollback false-positive and the common near-prompt case both still pass. Zero deps.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,19 @@ All notable changes to this project are documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.5.1] - 2026-06-30
+
+**Upgrade if you installed `0.5.0` from npm.** The `0.5.0` npm artifact was built
+before #29 was merged and shipped without the usage-retry anti-spam fix. `0.5.1`
+includes it. (The git tag `v0.5.0` already contained #29; only the npm tarball was
+behind.)
+
+### Fixed
+- Stop the usage-retry path from spamming an already-resumed session: a lingering
+  limit banner in scrollback no longer re-injects `Continue…` every poll. Detection
+  is now anchored to the live tail, and an `isWorking` gate stops the moment Claude
+  resumes (#29).
+
 ## [0.5.0] - 2026-06-30
 
 This release rolls up everything merged since `0.2.2`, including the API
@@ -18,10 +31,6 @@ overload backoff engine and interactive `/rate-limit-options` menu navigation.
 - Enable mouse scroll and vi copy-mode in tmux sessions created by the tool (#25).
 
 ### Fixed
-- Stop the usage-retry path from spamming an already-resumed session: a lingering
-  limit banner in scrollback no longer re-injects `Continue…` every poll. Detection
-  is now anchored to the live tail, and an `isWorking` gate stops the moment Claude
-  resumes (#29).
 - Require Claude to be in the foreground before driving the
   `/rate-limit-options` menu, preventing keystrokes from leaking into the wrong
   pane (#28).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   to the `API Error` render (mentioning the phrases in conversation can't
   trigger it), and the retry budget is kept across working ticks so a sticky
   flag stays bounded (#33).
+- tmux status bar indicator: the monitor now writes a per-pane status snapshot to
+  `~/.claude-auto-retry/status/<pane>.json` on every poll tick, and a new
+  `claude-auto-retry-tmux-status` script renders it as a status-bar segment
+  (`🟢AR` monitoring, `⏳AR 1h30m` waiting on a reset, `🟠AR 45s` overload backoff,
+  `🔴AR` gave up). Dependency-free POSIX shell; hides itself if a pane has no
+  monitor or its status file is stale (staleness is derived from the monitor's
+  actual poll interval, not a fixed constant).
 
 ### Fixed
 - `rate_limit` StopFailure events are no longer routed through the seconds-scale
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,35 @@
+# Changelog
+
+All notable changes to this project are documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [0.5.0] - 2026-06-30
+
+This release rolls up everything merged since `0.2.2`, including the API
+overload backoff engine and interactive `/rate-limit-options` menu navigation.
+
+### Added
+- Detect sustained API overload (`529`/`500`/`503`) and retry with exponential
+  backoff, including an event-driven (`StopFailure`) mode (#20, hardened).
+- Interactive navigation of the `/rate-limit-options` menu, driving it to
+  "Stop and wait" across any menu layout (#19, #26).
+- Enable mouse scroll and vi copy-mode in tmux sessions created by the tool (#25).
+
+### Fixed
+- Stop the usage-retry path from spamming an already-resumed session: a lingering
+  limit banner in scrollback no longer re-injects `Continue…` every poll. Detection
+  is now anchored to the live tail, and an `isWorking` gate stops the moment Claude
+  resumes (#29).
+- Require Claude to be in the foreground before driving the
+  `/rate-limit-options` menu, preventing keystrokes from leaking into the wrong
+  pane (#28).
+- Reliable retry submission plus session/weekly rate-limit detection (#7, #15, #22).
+- Correct an off-by-a-day wait when parsing reset times in offset timezones (#6, #23).
+- Unalias `claude` before defining the wrapper, fixing a zsh/bash `source` error (#10, #24).
+- Skip send-keys correctly when the foreground process is the shell, not Claude (#1).
+
+## [0.2.2] - 2026-03-31
+
+- Last published baseline release.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,23 @@ All notable changes to this project are documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+- Safeguard/AUP false-positive auto-retry: when the model's safeguards flag a
+  message ("safeguards flagged this message"), re-send a short retry up to
+  `safeguard.maxRetries` times, then give up loudly once. Detection is anchored
+  to the `API Error` render (mentioning the phrases in conversation can't
+  trigger it), and the retry budget is kept across working ticks so a sticky
+  flag stays bounded (#33).
+
+### Fixed
+- `rate_limit` StopFailure events are no longer routed through the seconds-scale
+  overload path — a session/usage limit is an hours-scale wait owned by the
+  usage path, and the misroute made the two fight (futile `Continue` retries
+  into a session-limited pane). The marker error type is validated at the
+  consumer too, so an outdated installed hook can't reintroduce it (#31).
+
 ## [0.5.1] - 2026-06-30
 
 **Upgrade if you installed `0.5.0` from npm.** The `0.5.0` npm artifact was built
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- **Chrome-aware detection.** Limit/overload/menu detectors now skip trailing UI
+  furniture (input box, footer, key hints, todo/task widget, status spinner,
+  `/usage-credits` hint) before reading the live tail, so a genuine banner behind a tall
+  task widget is still detected (fixes a ~54-min stall) while a banner merely quoted in
+  scrollback is not (#34).
 - Safeguard/AUP false-positive auto-retry: when the model's safeguards flag a
   message ("safeguards flagged this message"), re-send a short retry up to
   `safeguard.maxRetries` times, then give up loudly once. Detection is anchored
@@ -23,11 +28,29 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   actual poll interval, not a fixed constant).
 
 ### Fixed
+- The `/usage-credits` backstop no longer reopens the scrollback false positive: it only
+  fires when the companion sits in the live region (nothing but chrome below it), so a
+  resumed session's stale banner+companion can't drive spurious retries or a ~24h wait (#34).
+- `isWorking` is chrome-aware, matching `isRateLimited`: a live "esc to interrupt" footer
+  pushed up by a chrome stack is no longer missed, so retry text can't land in a
+  mid-flight session (#34).
+- The `/rate-limit-options` menu detectors are chrome-aware too, so a live menu behind a
+  widget is driven to "Stop and wait" instead of skipped (which risked confirming
+  "Upgrade your plan") (#34).
+- Overload detection is bounded to a max raw distance from the prompt, so the deeper
+  50-line capture can't reach an old quoted `API Error` buried behind a tall widget (#34).
+- Chrome classifiers are anchored to real footer/widget renders (pipe-anchored version,
+  indented task items, `⏵⏵` mode footer), so ordinary content — `Press ctrl+c…`, a
+  `→` rename, a flush-left `✓ …` summary, `Released v0.5.1` — is no longer stripped (#34).
 - `rate_limit` StopFailure events are no longer routed through the seconds-scale
   overload path — a session/usage limit is an hours-scale wait owned by the
   usage path, and the misroute made the two fight (futile `Continue` retries
   into a session-limited pane). The marker error type is validated at the
   consumer too, so an outdated installed hook can't reintroduce it (#31).
+
+### Changed
+- `customPatterns` are matched against the raw last-N lines (unchanged from pre-#34
+  semantics), not the chrome-skipped view — the user owns their own tradeoff (#34).
 
 ## [0.5.1] - 2026-06-30
 
--- a/DESIGN-NOTES.md
+++ b/DESIGN-NOTES.md
@@ -1,0 +1,201 @@
+# Design notes
+
+Forward-looking design notes for `claude-auto-retry`, grounded in a research pass
+(2026-06) against the installed Claude Code binary (v2.1.195, decompiled), the
+Anthropic API error schema, and Claude Code's hooks/transcript surfaces. Ordered by
+leverage. Items marked **[done]** ship in the current version; the rest are proposals
+with enough detail to execute.
+
+## 0. The core problem: we scrape a human-facing render
+
+The overload path detects a *terminal* API error by scraping `tmux capture-pane` and
+string-matching. That is fundamentally a heuristic over output meant for humans, and
+every false positive we've hit (a `res.status(503)` under edit, an HTTP status code in
+tool output in scrollback) and every false negative we worry about is a symptom of that
+layer choice.
+The hardening below makes the scraper as good as a scraper can be — but the strategic
+move is to stop using the scrape as the *trigger*.
+
+### Verified ground truth (informs everything else)
+
+- **Two render forms, opposite meaning.** Terminal (retries exhausted):
+  `API Error: <code> <body>` (colon form). Transient (still retrying):
+  `API Error (<code> …) · Retrying in 5s · attempt 3/10` (parens, and/or a `· Retrying`
+  suffix on the colon form). Acting on the transient form interrupts Claude's *own*
+  backoff. **[done]** — we require the colon form and suppress the retry suffix via the
+  working gate.
+- **Error → HTTP status → retryability.** Retryable: `429 rate_limit_error`,
+  `500 api_error`, `504 timeout_error`, `529 overloaded_error`, plus edge `502/503`
+  (no JSON body — plain text/HTML from Cloudflare/envoy, e.g. `503 no healthy upstream`).
+  Never retry: `400/401/402/403/404/413`. The official SDK retry set is `408/409/429 +
+  all 5xx`. **[done]** — default patterns cover `429|500|502|503|504|529` + `overloaded_error`.
+- **API-429 has no 3-digit code in the slot.** It renders
+  `API Error: Server is temporarily limiting requests (not your usage limit) · Rate
+  limited`. A code-keyed matcher misses it. **[done]** — phrase pattern added.
+- **529 is global capacity, not per-key.** Naive fixed backoff makes every client
+  resynchronize into the same overload window → use **full jitter**. See §3.
+- **Streaming caveat.** An error after a 200 (SSE `error` event) bypasses HTTP status.
+  Claude streams, so a mid-response overload may render differently than a pre-flight
+  failure. Not currently special-cased; low observed frequency.
+
+## 1. Replace the trigger with `StopFailure` (highest leverage) — **[done, v0.4.0]**
+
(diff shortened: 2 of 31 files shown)


cheapestinference (changes requested):
I verified the top findings by running this branch's `patterns.js`/`monitor.js` against fixtures and comparing with master (PR ⇒ fires / master ⇒ doesn't). The problem is real, the content-distance idea is elegant, and the observed 54-min stall is worth fixing — but as implemented the PR reopens the scrollback false positive it explicitly set out to preserve, in three independent ways, and creates dangerous asymmetries with the sibling detectors that still use the raw tail. Most-severe first.

**1. The `/usage-credits` backstop reopens the scrollback false positive — CONFIRMED end-to-end** (`src/patterns.js:116`)
It scans the last `max(tailLines*3, 40)` **raw** lines with no "real work below" discipline, bypassing the contentTail it coexists with. Reproduced through the branch's actual `processOneTick`: stale banner + companion line in scrollback of an *already-resumed* session → up to `maxRetries` (5) `Continue where you left off…` injections 30s apart, and — because the stale `resets 2am` has passed — `calculateWaitMs` rolls to *tomorrow*, re-entering `waiting` for up to ~24h. It takes **38 lines** of new output to escape the window. The PR's own negative test ("quoted banner with real work below") passes only because its fixture omits the companion line — which the live render always prints, so post-resume scrollback always contains it. The backstop needs the same liveness discipline as the main path (e.g. require no non-chrome content *below* the companion), or it should go.

**2. Chrome classifiers match ordinary content — CONFIRMED with 5 probes** (`src/patterns.js:40,41,63`)
Each of these real output lines is classified as chrome and stripped, pulling a stale banner 14+ raw lines up back into the 12-content-line window (PR ⇒ true, master ⇒ false): `Press ctrl+c to stop the dev server`, `⎿ Renamed a.js → b.js` (arrow), `✓ Fixed the bug` (extremely common summary bullet), `Released v0.5.1` (any semver), `Run /rc to reconnect`. Unanchored single glyphs (`↑↓←→`, `ctrl+`, `✓✔`, `\bv\d+\.\d+\.\d+\b`) are too loose for a strip-list whose miss cost is a false retry; they need anchoring to the actual footer/widget renders (position, leading whitespace, full-line shape).

**3. `isWorking` stayed raw-tail, defeating the #29 anti-spam guard — CONFIRMED** (`src/patterns.js:264`)
With this PR's own test-fixture chrome stack (~15 lines), a live `✻ … (esc to interrupt)` above it is invisible to `rawTail(12)` while chrome-aware `isRateLimited` sees the lingering banner → `!isRateLimited || isWorking` at `monitor.js:149` fails both ways and the waiting branch sends retry text into a mid-flight session. The rawTail exemption is well-intentioned (the comment explains it), but it means the two functions measure *different bottoms of the same pane*. `isWorking` needs the same chrome treatment with the working footer excluded from the strip-list (it already deliberately isn't in `CHROME_LINE`), i.e. it can use contentTail safely — the footer is content by construction.

**4. `customPatterns` semantics silently changed, both directions — CONFIRMED** (`src/patterns.js:107`)
User regexes now test the chrome-stripped window: a pattern keyed on footer text (e.g. usage-percent) stops firing (master true → PR false), and patterns can now reach stale quoted text deeper in scrollback (master false → PR true). At minimum a README/CHANGELOG note; arguably custom patterns should keep the raw window since the user controls their own false-positive tradeoff.

**5. Menu detectors still raw-tail: retry can be typed into an open menu — PLAUSIBLE** (`src/patterns.js:160`)
`isRateLimitOptionsPrompt`/`menuStepsToWaitOption` kept `slice(-tailLines)`. Function-level repro: live menu pushed up by a tall widget → prompt = false, `isRateLimited` = true → menu branch skipped, usage-wait entered, and the later `sendKeys` types text + Enter into the open menu, where Enter confirms the highlighted default — "Upgrade your plan". Held at PLAUSIBLE only because it's unverified that the real TUI ever renders a tall widget below the live menu. If chrome-awareness lands, it must land on all four detectors at once.

**6. Overload detectors + 50-line capture reach much older stale errors — CONFIRMED, partly by design** (`src/patterns.js:220`, `src/monitor.js:97`)
A quoted `API Error: 529 …` that ended a turn (only blanks/chrome below) now triggers `detectOverload` from up to ~50 raw lines up, where master's 20-line capture couldn't even see it. "Error with only chrome below" *is* the terminal-error render, so some of this is intended — but the 2.5× deeper capture widens the quoted-log case considerably. Worth a bounded content-distance for overload too (e.g. cap how far above the prompt the error may sit).

Cleanup — CONFIRMED, non-blocking: `/\/usage-credits\b/i` exists twice (`CHROME_LINE` and `USAGE_CREDITS`) with no shared constant, and the `OVERLOAD_TAIL_LINES` comment still describes "the 20-line capture" that is now 50.

romerjon:
Thanks — every finding reproduced and fixed test-first (a failing test driving the real `patterns.js`/`processOneTick`, then the fix). Suite is green (215 tests). One commit per finding.

**1 — `/usage-credits` backstop reopened the false positive.** The backstop now has the main path's liveness discipline: it locates the last companion line and only fires when `all.slice(idx+1).every(isChromeLine)` — i.e. nothing but chrome below it. A resumed session's stale banner+companion (real output rendered below) no longer triggers; the "banner behind a recognized widget, only chrome below" case still does. (`f602d9e`)

**2 — chrome classifiers matched ordinary content.** Anchored to real renders instead of bare glyphs: version is pipe-anchored (`\|\s*v\d+\.\d+\.\d+`, so `Released v0.5.1` is content); task items require leading indentation (a flush-left `✓ Fixed the bug` is content); dropped the loose `ctrl+ / ↑↓←→ / /rc` matcher and replaced it with the anchored `^⏵⏵` mode footer + `shift+tab to cycle/select` + `? for shortcuts`. All 5 probes (`Press ctrl+c…`, `⎿ Renamed a.js → b.js`, `✓ Fixed the bug`, `Released v0.5.1`, `Run /rc to reconnect`) now stay content. (`40f982c`)

**3 — `isWorking` stayed raw-tail.** Now uses `contentTail`, so it measures the same bottom as `isRateLimited`. To keep it safe, `isChromeLine` excludes working lines (`WORKING_PATTERNS` moved up and reused), so a live `✻ … (esc to interrupt)` footer is never stripped as furniture even though it matches the spinner glyph. The idle `✻ Brewed for …` spinner is still chrome. (`bbb62c9`)

**4 — `customPatterns` semantics.** Restored to master's raw-tail window (`all.slice(-tailLines)`), so footer-keyed patterns fire again and can't reach deeper than before. Documented in README + CHANGELOG. (`ef21cdd`)

**5 — menu detectors still raw-tail.** `isRateLimitOptionsPrompt` and `menuStepsToWaitOption` are chrome-aware now too (menu lines aren't chrome, so `contentTail` keeps them). All four detectors share one window; a menu behind a tall widget is driven to "Stop and wait" instead of skipped. A menu only quoted above live work is still ignored. (`22de34f`)

**6 — overload + 50-line capture.** Added a raw-distance bound: `contentTail` takes an optional `maxRaw`, and the overload `tail()` passes `OVERLOAD_MAX_RAW_LINES=20` (master's original depth). An `API Error` reachable only by chrome-stripping past a tall widget stays out; a terminal error just above the input box still fires. The reset-anchored limit path stays unbounded by design. (`96b3c18`)

**Cleanup** (`dbb8ec3`): shared `USAGE_CREDITS` constant (was duplicated in `CHROME_LINE` + backstop); fixed the `OVERLOAD_TAIL_LINES` comment that still said "20-line capture".

cheapestinference:
Thanks for the thorough rework — the false-positive discipline here is solid, and I couldn't break the `/usage-credits` backstop or `isWorking`. Two things to address before merge, one functional and one integration:

### 1. The flagship fix doesn't fire for the boxed input render

`contentTail` strips only the *contiguous* trailing chrome block and stops at the first line it doesn't classify as chrome. Claude Code's input box middle line renders as `│ >                        │`, which isn't matched by any `CHROME_LINE` entry:

- the box-only rule `/^[\s─│╭╮╰╯…|]+$/` fails because of the `>`
- the empty-prompt rule `/^\s*[❯>]\s*$/` fails because the line starts with `│`, not whitespace

Since the todo/task widget renders *above* the input box, the strip halts at `│ > ` and never reaches the widget or the banner above it. All the tests model the input as a bare `❯ ` line between two rules, so they don't exercise this — but the repo itself treats the boxed form as a real render (`test/overload.test.js:107`, `isWorking('│ > ')`).

Reproduction against this branch's `patterns.js`:

```js
const widget = ['  8 tasks (4 done, 1 in progress, 3 open)',
  '  □ a','  □ b','  □ c','  □ d','  □ e','  □ f','  □ g','   … +3 completed',
  '  new task? /clear to save 300k tokens'];
const banner = "You've hit your session limit · resets 3pm (UTC)";

const bare  = ['───────','❯ ','───────','  ⏵⏵ auto mode on'];
const boxed = ['╭────────────────────────╮','│ >                      │','╰────────────────────────╯','  ? for shortcuts'];

isRateLimited([banner, ...widget, ...bare ].join('\n'), [], 12); // → true  ✅
isRateLimited([banner, ...widget, ...boxed].join('\n'), [], 12); // → false ❌ (banner behind widget, boxed input)
```

The direction is conservative (it never over-reaches into scrollback, so no new false positives) — it just doesn't deliver the advertised "banner behind a tall widget" detection for the render Claude Code actually uses. Fix is to classify the box's own lines as chrome (the middle `│ … │` row and the `╰──╯` bottom), plus a test that models the boxed input above a widget.

### 2. Branch is based on pre-#33 master and will revert safeguard if merged naively

The merge-base predates #31/#33, so this branch's `patterns.js`/`monitor.js` contain no `safeguard` code at all. Since the PR rewrites those two files, a merge that takes the branch's versions silently drops the safeguard detector/retry loop that shipped in #33. It *is* resolvable — keeping both blocks (your `isWorking` comment + master's `safeguardMatch`/`detectSafeguard`) merges cleanly and the full suite stays green (241 tests, incl. the 24 safeguard ones), and `safeguardMatch` then inherits the chrome-aware `tail()` for free. Please rebase onto current master so the resolution is explicit in the PR rather than left to merge time.

romerjon:
Both addressed.

**1. Boxed input render `│ > │` (`e022ece`).** Confirmed your repro and fixed it — the strip halted at the input box middle row because the all-box-chars rule breaks on the `>`/typed text and the empty-prompt rule needs a leading whitespace. Added a `│ … │` row to `CHROME_LINE` (U+2502, not ASCII `|`, so markdown `| tables |` are unaffected; and `contentTail` only strips *trailing* chrome, so a mid-pane box row with content below is untouched). Your exact bare-vs-boxed cases are now tests:

```
bare  → true
boxed → true   (was false)
```

**2. Rebased onto current master.** The safeguard-coexistence resolution is now explicit in the branch rather than left to merge time. `safeguardMatch` picks up the chrome-aware `tail()` for free (and inherits the Finding 6 raw-distance bound). Also merged cleanly with the #30 tmux-status work that landed in the meantime — full suite is **265 green** (incl. the safeguard + status-file tests). CHANGELOG Unreleased now carries #34 + #33 + #30 + #31 together.

One note for whoever merges: I found a *separate* pre-existing bug while running the branch on live sessions — `calculateWaitMs` rolls **any** past reset time a full day forward, so a banner detected minutes after its reset (which the chrome-aware guard makes more likely — it holds detection until the session settles) parks ~24h. I've got a standalone fix + tests for it and will open it as its own PR so it doesn't muddy this one.

romerjon:
Heads-up: I pushed one more commit (`fe45ef5`) — a **new finding from running this branch live**, not part of your two requested changes. Flagging it explicitly so you can accept or split it.

**What I saw:** a session hit its session limit, dispatched a background agent, and kept making progress — but `isWorking` returned false for every state it was actually in (`✻ Waiting for 1 background agent to finish`, `Backgrounded agent (↓ to manage …)`). With the stale `resets 3pm` banner still lingering just above that activity, the waiting branch re-injected `Continue` into the working session, cycle after cycle. (On my box this rode on a separate `calculateWaitMs` grace-window fix — PR #35 — which turned what used to be a 24h stall into fast 60s re-cycles, making the spam visible.)

**The commit:**
1. Adds background-agent states to `WORKING_PATTERNS`, so the waiting-path `isWorking` gate fires (no injection) and — since working lines are excluded from `CHROME_LINE` — a banner rendered above them stays out of the live window.
2. Gates the *monitoring*-path usage-limit detection on `!isWorking` as well: a session that's actively working can't be blocked by a live limit, so a banner visible while it works is stale — don't even enter a wait (otherwise it re-detects every poll). This mirrors the discipline the scraper-overload path already uses.

Test-first (background-agent `isWorking` cases + two `processOneTick` regressions), full suite green.

It's in the `isWorking` mechanism this PR reworks, so it felt like the natural home — but if you'd rather land #34 as-is against your two items and take this as a follow-up PR, say the word and I'll pull `fe45ef5` back out into its own branch.

cheapestinference:
Verified both new commits against the real exports and `processOneTick`. `e022ece9` fixes my boxed-input finding and the tests genuinely pin it (reverting just the rule fails exactly the two new tests). But the rule as written is broader than the input box, and `fe45ef54` has a demonstrated failure mode that's worse than the bug it fixes. Details:

### 1. `e022ece9` — the `│ … │` rule matches more than the input row

`/^\s*│.*│\s*$/` classifies **any** U+2502-bordered row as chrome, not just the input box. Two consequences:

**False retry (confirmed):** tool output with unicode box borders as the last thing on screen — psql/mysql unicode-border tables, `duf`, any boxed CLI — now strips as chrome, pulling a stale scrolled-past banner back into the detection window:

```
You've hit your session limit · resets 3pm (UTC)   <- stale, scrolled past
● Bash(psql -c "select * from users limit 8")
  ⎿  ┌────────┬───────────┐
     │ 0      │ user0     │        <- all rows match the new rule
     └────────┴───────────┘
❯
```

Pre-fix: `isRateLimited` false (table rows were content). Post-fix: **true** → false retry into a session that already moved on. ASCII `|` markdown tables are safe as you noted — the gap is U+2502 *output*, not tables in prose. The same class loosens the `/usage-credits` backstop's `every(isChromeLine)` liveness gate.

**Missed detection (realism unconfirmed — you have the live captures):** if Claude Code ever renders the limit banner, or the `/rate-limit-options` menu, inside a `╭─╮ │ │ ╰─╯` box at the tail, the new rule strips the banner/menu rows themselves. The menu case is the nasty one: `isRateLimitOptionsPrompt` flips to false while `isRateLimited` above stays true, so the monitor enters a wait and later sends text+Enter into an open menu whose highlighted default is "Upgrade your plan" — the exact hazard this PR's own menu handling exists to prevent. If your captures show these renders are never boxed, fine; otherwise this needs the tighter rule regardless.

**Suggested tightening:** anchor to the prompt row specifically — `/^\s*│\s*[>❯].*│\s*$/` (the `│ > │` empty row and `│ > typed text │` both carry the prompt glyph). Caveat to check against a live capture: continuation rows of multi-line typed input (`│   second line │`, no glyph) — if those render, a two-row allowance below a glyph row, or matching only above the footer, may be needed. Either way the discriminator should be the prompt glyph, not "anything between two bars".

### 2. `fe45ef54` — please split, and rework the gate before it lands anywhere

The live bug you saw is real, but this fix suppresses legitimate retries indefinitely. `Backgrounded agent (↓ to manage · ctrl+o to expand)` is a **transcript** line — it lingers in scrollback after the agent finishes. Repro through the real `processOneTick`:

```
● Task(build the parser)
  ⎿  Backgrounded agent (↓ to manage · ctrl+o to expand)   <- agent long finished
● Done. The parser passes all 14 tests.
You've hit your session limit · resets 3pm (Europe/Zurich) <- LIVE banner
❯
```

`isWorking` = true forever (the pane is static while idle), so the monitoring tick never enters a wait, and a waiting-state tick returns `user-continued` without sending — every tick, indefinitely. The tool silently never retries a genuinely limited, idle session. `/\bbackgrounded agent\b/i` also matches ordinary prose ("the backgrounded agent approach avoids blocking") within the last 12 content lines.

Two observations for the split-out version:

- The `✻ Waiting for N background agents to finish` pattern alone already fixes your observed cycle both ways (waiting-branch `|| isWorking` gate, plus working lines keeping the stale banner out of the chrome window) — and that render is live-only, so it's safe. Keep that; drop `\bbackgrounded agent\b` or anchor it to the full live render and accept the consequences knowingly.
- The monitoring-path `&& !isWorking(stripped)` gate changes detection semantics for **every** working pattern, widening any over-match from "skip one injection" to "never detect the limit at all". That's the piece that turns a fuzzy pattern into total feature failure, and it deserves its own justification in the split PR.

`e022ece9` (with the tightened regex) is the only part of the two commits that belongs in this PR. Suite is green at HEAD (269/269) and the rebase onto master checks out — findings 1–8 from the previous round all hold as reworked, so once the regex is tightened and `fe45ef54` is split out, this is mergeable.

romerjon:
Both done.

**Boxed-input regex tightened** (amended into the boxed-input commit, now `ddb7b0d`): `/^\s*│\s*[>❯].*│\s*$/` — anchored to the prompt glyph, not "anything between two bars". Your psql repro is now a test: a unicode-border table above a stale banner stays content (banner excluded → no false retry), while bare `❯` and boxed `│ > │` inputs above a widget both still detect. I left the multi-line-typed-input continuation-row case (`│  second line │`, no glyph) as-is for now — if your live captures show that render, I'll add a below-a-glyph-row allowance; happy to take a capture if you have one.

**`fe45ef54` split out** — you're right, `Backgrounded agent` is a transcript line and gluing it into `isWorking` makes the tool never retry a genuinely-limited idle session. Dropped from this PR. I'm opening it as its own PR with only the live-only `Waiting for N background agents` pattern and a separate justification for the monitoring-path gate. This PR is back to just findings 1–8 + the tightened boxed-input rule (266 green).

cheapestinference (changes requested):
The over-broad box rule is genuinely fixed — the boxed-input chrome rule is now anchored to the prompt glyph (`/^\s*│\s*[>❯].*│\s*$/`), so a unicode-border table row (`│ 0 │ user0 │`, no glyph) no longer strips a stale banner sitting above it, while a real limit/error banner (unboxed, or a warn-glyph box) is still detected. Verified adversarially against the real `isRateLimited` export. 266/266.

Two things before this can land, though:

**1. The "fe45ef5 amended into ddb7b0d" note isn't accurate for this PR.** `git show ddb7b0d` is only the boxed-input fix (`patterns.js` + 4 `patterns.test.js` cases). fe45ef5's actual content — the `waiting for N background agents` `WORKING_PATTERNS` entry and the `&& !isWorking` gate on the monitoring path — is not on this branch; it's in #36 (`8c24f35`). `git merge-base --is-ancestor fe45ef5 pr-34` → no.

**2. As a self-contained PR, #34 regresses.** It adds `Backgrounded agent` to `CHROME_LINE` (`src/patterns.js:79`), but the monitoring path at `src/monitor.js:359` is still `if (isRateLimited(...))` — *not* gated on `!isWorking`. On a working session with a stale `resets 3pm` banner above a running background agent, the background lines strip as chrome and pull the banner into the tail → `isRateLimited=true`, `isWorking=false` → `enterUsageWait` fires on a live session. That's exactly the retry-injection spam fe45ef5 was written to prevent, and its compensating gate lives in #36, not here. (Confirmed empirically against the real exports and `processOneTick`.)

So either land #36's content together with / before this one (it carries both the `!isWorking` gate and the `waiting for N background agents` working pattern), or drop the `Backgrounded agent` `CHROME_LINE` addition from #34. Whichever you choose, worth correcting the "amended" note so the intended merge order is explicit on the PR — right now #34 alone puts the regression on master.

romerjon:
You're right on both counts — thanks for catching the note error.

**1. Corrected:** `fe45ef5` is *not* in this PR. My "amended into ddb7b0d" note was wrong — `ddb7b0d` is only the boxed-input fix. The `waiting for N background agents` pattern + the `!isWorking` monitoring gate live in **#36** (`8c24f35`). Apologies for the misleading note.

**2. You're right that #34 regresses in isolation, and it's not separable.** #34 makes `Backgrounded agent` chrome (so a stale banner behind a *finished*-agent transcript is still detected and retried — the idle-limited case), but that's only safe together with #36's `!isWorking` gate (so a *running*-agent session isn't spammed). The two are a genuine unit: #34 = the chrome-aware classification, #36 = the gate that makes it safe. There's no version of #34 that's both self-contained and correct — dropping `Backgrounded agent` from `CHROME_LINE` would just trade the spam-regression for a miss-regression (an idle limited session with a lingering transcript would stop being detected, which master handles today).

Since **#36 is stacked on #34 and already approved**, merging #36 brings both correctly (its diff = #34's commits + the gate). My suggestion: **merge #36** and close #34 as subsumed — or merge #34 immediately followed by #36 if you prefer to keep the two-PR history. Happy to **fold #36 back into #34** as a single self-contained PR if that's cleaner for you; just say which and I'll restructure. What I'd avoid is merging #34 alone, exactly as you flagged.

romerjon:
Two follow-up-review fixes (`bcf9943`): **(F3)** the `/usage-credits` backstop now requires a *reset* line nearby, not just a LIMIT match, so a session merely explaining usage limits (companion + loose `usage limit` text, no reset) no longer false-fires. **(F4a)** the boxed-input rule's `.*` → `[^│]*`, so a unicode table row whose first cell is `>` (`│ > │ meaning │`) isn't stripped as the input box — closes the residual on your psql concern. Regressions added.

romerjon:
Follow-up-review fix (`7db0c8c`): tightened four `CHROME_LINE` entries that still matched ordinary **prose** as a last-content-line. Same false-positive class as Finding 2 (a stale banner sitting exactly at the content boundary gets stripped back into the window), so anchored each to how the furniture actually renders:

- **auto mode** — dropped the bare `/\bauto mode\b/i` (matched "auto mode is enabled in your settings"). The footer is already covered by `/^\s*⏵⏵/`; the permission notice now anchors to `/Allowed by auto mode/i`.
- **task header** — `/^\s*\d+\s+tasks?\b/i` → `/^\s*\d+\s+tasks?\s+\(/i`. The `(` count is part of the real widget header (`8 tasks (4 done, …)`), so "3 tasks remain in the backlog" is no longer stripped.
- **backgrounded agent** — `/Backgrounded agent|to manage · /i` → `/Backgrounded agent \(|to manage · /i`. The notice carries the `(` (or `to manage ·`), so "Backgrounded agent finished the lint run" is no longer stripped.
- **save hint** — `/new task\?|\/clear to save/i` → `/\/clear to save/i`. The standalone "new task?" matched prose questions ("Should I start the new task?").

Test-first: `F4B_PROSE_PROBES` reproduces each false strip through the real `isRateLimited` (stale banner, 13 prose copies, input box, tail=12 → was `true`, now `false`); `F4B_CHROME_RENDERS` guards that the genuine renders still strip (banner behind them still detected). One existing companion-backstop fixture used a bare `8 tasks` header (no `(`) as incidental furniture below the companion — updated to the real `8 tasks (4 done, …)` render. **276/276.**

(#36 rebased onto this — `aa16115`, 281/281.)

cheapestinference:
Superseded by #43. I rewrote `master`'s history to clean up commit metadata, which changed the base commits this branch was built on and invalidated the diff here. Recreated as #43 — rebased onto the current `master`, **your commits and authorship preserved**, tests green, review carried over. Nothing needed from you; continue any follow-up on #43. Thanks again!

PR #35, opened 2026-07-07
fix(time-parser): grace window so a just-passed reset retries promptly, not in ~24h

romerjon:
## Problem

`calculateWaitMs` rolls **any** past reset time a full day forward:

```js
let diff = getTargetTimestamp(parsed.hour, parsed.minute) - now.getTime();
if (diff < 0) diff += 86400_000; // tomorrow
```

That's correct for a reset genuinely hours in the past, but wrong at the reset *boundary*. The monitor can settle on a limit banner minutes-to-an-hour **after** its reset — the chrome-aware anti-spam guard deliberately holds detection until the session stops working, which can push the first detection just past the reset. `diff < 0` then parks the session ~24h even though the limit has effectively cleared.

Observed live: `You've hit your session limit · resets 10am (Europe/Zurich)` detected at 10:03 Zurich → **86273s (~24h)** wait, on a box whose sessions were otherwise fine to continue right after 10am.

## Fix

A 1-hour grace window (`rollPastReset`): a reset within the last hour is treated as just-reset (`diff → 0`, so the wait is only the margin — it retries promptly and self-corrects if still limited); only a reset more than an hour past rolls to tomorrow. Applied to both the exact and ambiguous (am/pm) branches.

```
resets 10am, detected 10:03 Zurich:  60s   (was 86273s)
resets 3pm,  detected 12:48 Zurich:  ~2.2h (unchanged — future reset)
```

## Tests

Test-first: two failing boundary cases (3 min and 55 min past → prompt retry) plus a guard that a reset well over an hour past still rolls to tomorrow. The #6 timezone regressions and all future-reset behavior are unchanged. Full suite green.

Self-contained — touches only `src/time-parser.js` and its test. Surfaced while running #34 on live sessions; kept separate so it doesn't muddy that PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,19 @@ All notable changes to this project are documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.5.1] - 2026-06-30
+
+**Upgrade if you installed `0.5.0` from npm.** The `0.5.0` npm artifact was built
+before #29 was merged and shipped without the usage-retry anti-spam fix. `0.5.1`
+includes it. (The git tag `v0.5.0` already contained #29; only the npm tarball was
+behind.)
+
+### Fixed
+- Stop the usage-retry path from spamming an already-resumed session: a lingering
+  limit banner in scrollback no longer re-injects `Continue…` every poll. Detection
+  is now anchored to the live tail, and an `isWorking` gate stops the moment Claude
+  resumes (#29).
+
 ## [0.5.0] - 2026-06-30
 
 This release rolls up everything merged since `0.2.2`, including the API
@@ -18,10 +31,6 @@ overload backoff engine and interactive `/rate-limit-options` menu navigation.
 - Enable mouse scroll and vi copy-mode in tmux sessions created by the tool (#25).
 
 ### Fixed
-- Stop the usage-retry path from spamming an already-resumed session: a lingering
-  limit banner in scrollback no longer re-injects `Continue…` every poll. Detection
-  is now anchored to the live tail, and an `isWorking` gate stops the moment Claude
-  resumes (#29).
 - Require Claude to be in the foreground before driving the
   `/rate-limit-options` menu, preventing keystrokes from leaking into the wrong
   pane (#28).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   to the `API Error` render (mentioning the phrases in conversation can't
   trigger it), and the retry budget is kept across working ticks so a sticky
   flag stays bounded (#33).
+- tmux status bar indicator: the monitor now writes a per-pane status snapshot to
+  `~/.claude-auto-retry/status/<pane>.json` on every poll tick, and a new
+  `claude-auto-retry-tmux-status` script renders it as a status-bar segment
+  (`🟢AR` monitoring, `⏳AR 1h30m` waiting on a reset, `🟠AR 45s` overload backoff,
+  `🔴AR` gave up). Dependency-free POSIX shell; hides itself if a pane has no
+  monitor or its status file is stale (staleness is derived from the monitor's
+  actual poll interval, not a fixed constant).
 
 ### Fixed
 - `rate_limit` StopFailure events are no longer routed through the seconds-scale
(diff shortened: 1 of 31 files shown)


cheapestinference:
Reviewed and reproduced the live stall — the exact-time branch fix is right, and the retry-burst risk of an over-eager grace window is already bounded by the monitor's existing machinery (`maxRetries` + 30s cooldowns + the `isWorking` gate), so a misfire self-corrects. The three tests pin the boundary well. One real issue in the ambiguous branch to fix before merge:

### Ambiguous both-past outside grace rolls to the *wrong* tomorrow occurrence

Master's else-branch rolled `d1` (the am interpretation — always the earlier of the two, hence the earliest next occurrence) to tomorrow. The PR rolls `Math.max(d1, d2)` — the **pm** interpretation — which lands 12h later than the earliest occurrence:

```
"resets 10 (Europe/Zurich)" detected at 23:30 Zurich (10pm passed 1.5h ago, outside grace):
  master:  10.52h  (10am tomorrow — earliest next occurrence)
  PR #35:  22.52h  (10pm tomorrow)
```

Verified both numbers by running the case against each branch. `Math.max` is the right input for the *grace check* (is the most recent interpretation just-passed?), but the wrong input for the *roll* (which next occurrence comes first?). Since `t1 < t2` always, outside the grace window this waits exactly 12h longer than master in every both-past case. Suggested shape:

```js
else {
  const recent = Math.max(d1, d2);
  // Grace on the most recent interpretation; otherwise the earliest occurrence tomorrow.
  target = recent > -RESET_GRACE_MS ? 0 : Math.min(d1, d2) + 86400_000;
}
```

(One could argue both-past should *always* retry promptly — a banner shows a future time when printed, so both interpretations being past means the detection is stale regardless — but keeping the conservative roll for >1h-past is fine as long as it targets the earliest occurrence.)

Please add the regression test for this case (e.g. `resets 10`, 23:30 Zurich → expect ~10.5h, not ~22.5h); the current suite only exercises the ambiguous branch's future paths, which is why this slipped through. The within-grace ambiguous case works correctly as-is (verified: `resets 10` at 22:30 Zurich → ~1min, vs master's 11.5h) — worth a test too since it's a behavioral win of this PR.

Everything else checks out: branch is based on current master, suite green (242 tests), the 1h window vs the #6 regression boundary (1h20m still rolls) is preserved, and the strict `>` boundary at exactly -1h is fine.

romerjon:
Fixed — great catch. The else-branch now grace-checks `Math.max(d1,d2)` (most recent interpretation) but rolls `Math.min(d1,d2) + 24h` (earliest occurrence), so `resets 10` at 23:30 Zurich targets 10am-tomorrow (~10.5h) instead of 10pm-tomorrow. Added both regressions you asked for — the outside-grace `~10.5h` case and the within-grace `22:30 → prompt` win. Suite green (244).

cheapestinference (dismissed):
Verified `de1fdd02` against the real `calculateWaitMs` with an injected `now`. The else-branch now grace-checks `Math.max(d1, d2)` (the most-recent interpretation) while rolling to `Math.min(d1, d2) + 24h` (the earliest occurrence) — which is exactly right:

- pm reset that passed 30–59 min ago → ~60s prompt retry (not the old ~24h).
- both-past outside grace ("resets 10" @ 23:30) → 10.5h to 10am tomorrow, not the old 22.5h pm bug.
- an ambiguous *future* reset can't grace-fire early: the else is only reached when both candidates are ≤ 0 (guarded by the three preceding `> 0` branches), so `Math.max` is always ≤ 0 there — a reset 5 min out returns ~6 min, a non-ambiguous 2h reset returns 2.02h.

Boundary sweep is monotonic and sane (59m → 60s, exactly 60m → roll, 61m → roll), and the exclusive `> -GRACE` boundary fails toward waiting. 244/244, including the two new ambiguous regression tests. LGTM.

romerjon:
Follow-up-review fix (`b09724e`): a 24h-style `resets 12:30` (no am/pm) parsed as ambiguous hour=12, and the ambiguous branch then built `new Date("…T24:30:00Z")` — Invalid → throws every tick → the monitor crash-loops to exit. Now the pm interpretation is `(hour+12) % 24` (12 → midnight, never 24), and `parseResetTime` rejects an out-of-range clock (e.g. a bare `resets 30`) → the safe fallback wait. Pre-existing on master, but it's in this branch's function so I folded the fix here; two regressions added.

cheapestinference:
Superseded by #44. I rewrote `master`'s history to clean up commit metadata, which changed the base commits this branch was built on and invalidated the diff here. Recreated as #44 — rebased onto the current `master`, **your commits and authorship preserved**, tests green, review carried over. Nothing needed from you; continue any follow-up on #44. Thanks again!

PR #36, opened 2026-07-07
fix: treat "Waiting for N background agents" as working (split from #34)

romerjon:
Split out of #34 at your request. **Stacked on #34** (`fix/limit-detection-chrome-aware`) since it builds on that PR's `isWorking`/`WORKING_PATTERNS` — until #34 merges, the diff here also shows #34's commits; the only new commit is the last one.

## The bug (live)

A session hit its session limit, dispatched/awaited a background agent, and kept making progress — but showed none of the streaming/retry footers, so `isWorking` read it as idle. With a stale `resets 3pm` banner lingering above the activity, the waiting branch re-injected `Continue` every grace-window cycle.

## The fix — narrower than my first attempt

You were right that gluing the `Backgrounded agent` **transcript** line into `isWorking` is a worse bug (it lingers → `isWorking` latches true → a genuinely-limited idle session never gets retried). So:

- **Only** the live-only `waiting for N background agents to finish` render goes into `WORKING_PATTERNS`. It vanishes the instant the agent finishes, so it can't pin `isWorking` on an idle limited pane. A finished-agent `Backgrounded agent` transcript is explicitly **not** working — there's a counter-repro test for exactly your scenario (`isWorking` false → the tool still retries).
- The monitoring-path `!isWorking` gate is included **with the justification you asked for**: it mirrors the waiting branch's existing `|| isWorking` guard and closes the re-detection cycle (the `Waiting for N` line alone stops injections, but a stale banner sitting <12 content lines above it still re-enters a wait every tick otherwise). The tradeoff you flagged is real — it widens every working pattern from "skip one injection" to "don't detect while it matches" — so with the `Backgrounded agent` line gone, every remaining working pattern is a **live-only** render by construction, bounding the blast radius. If you'd still rather drop the monitoring-path gate and accept the cosmetic re-cycle, say so and I'll pull it.

Test-first: `Waiting for N` → working; finished-agent transcript → not working and still retried; monitoring tick doesn't enter a wait while a background agent runs. Full suite green (270).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,19 @@ All notable changes to this project are documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.5.1] - 2026-06-30
+
+**Upgrade if you installed `0.5.0` from npm.** The `0.5.0` npm artifact was built
+before #29 was merged and shipped without the usage-retry anti-spam fix. `0.5.1`
+includes it. (The git tag `v0.5.0` already contained #29; only the npm tarball was
+behind.)
+
+### Fixed
+- Stop the usage-retry path from spamming an already-resumed session: a lingering
+  limit banner in scrollback no longer re-injects `Continue…` every poll. Detection
+  is now anchored to the live tail, and an `isWorking` gate stops the moment Claude
+  resumes (#29).
+
 ## [0.5.0] - 2026-06-30
 
 This release rolls up everything merged since `0.2.2`, including the API
@@ -18,10 +31,6 @@ overload backoff engine and interactive `/rate-limit-options` menu navigation.
 - Enable mouse scroll and vi copy-mode in tmux sessions created by the tool (#25).
 
 ### Fixed
-- Stop the usage-retry path from spamming an already-resumed session: a lingering
-  limit banner in scrollback no longer re-injects `Continue…` every poll. Detection
-  is now anchored to the live tail, and an `isWorking` gate stops the moment Claude
-  resumes (#29).
 - Require Claude to be in the foreground before driving the
   `/rate-limit-options` menu, preventing keystrokes from leaking into the wrong
   pane (#28).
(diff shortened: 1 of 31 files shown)


cheapestinference (dismissed):
Reviewed the one commit that's new vs #34 (`8c24f35`). The `waiting for \d+ background agents? to finish` `WORKING_PATTERNS` entry is correct and specific — verified against the real exports across singular/plural, 0, uppercase, large N, and buried-under-chrome, with no false positives against any limit/overload/safeguard render. Precedence is right too: the monitoring gate becomes `isRateLimited(...) && !isWorking(...)` (`monitor.js:367`), so a live background-agent pane with a stale banner above it defers instead of injecting, and the moment the agent finishes (only the `Backgrounded agent` scrollback notice remains) a genuinely-limited idle pane is retried again. 270/270.

The safety argument rests on "waiting for N background agents to finish" being a live-only render (it clears the instant the agent finishes) — which you've documented in both the commit message and the `:367` comment, and the worst case is bounded. Reasonable, and consistent with the existing `esc to interrupt` / `Retrying in` patterns' exposure.

One process note: this is stacked on #34, so it needs #34's commits to land first — and per my #34 review, #34 shouldn't merge on its own (without this commit it carries a retry-injection regression, since the `!isWorking` gate and the working pattern both live here). Cleanest is probably to land these together, or merge #36 and close #34 as superseded — flagging so the merge order is deliberate rather than #34-then-a-gap-then-#36.

romerjon:
Follow-up adversarial-review fix (`1ac28c1`): **dropped the monitoring-path `!isWorking` gate** — you were right that it was risky. It's not just theoretical: `Retrying in …`/`attempt N/M` match *transcript* text (a flaky deploy/test log lingering in the tail), so with the gate a genuinely-limited session with such a line would never be retried. The waiting branch's `|| isWorking` guard already stops injection into a working session (which prevents the background-agent spam this PR targets); the only cost of dropping the gate is a cosmetic re-detection cycle that never injects. Added a regression. Rebased onto the updated #34.

romerjon:
Rebased onto the updated #34 (`aa16115`) after the F4b chrome-prose-anchoring fix landed there — the two commits on top of #34 are unchanged (background-agent working pattern + dropped monitoring-path gate). No conflicts, **281/281**.

cheapestinference:
Superseded by #45. I rewrote `master`'s history to clean up commit metadata, which changed the base commits this branch was built on and invalidated the diff here. Recreated as #45 — rebased onto the current `master`, **your commits and authorship preserved**, tests green, review carried over. Nothing needed from you; continue any follow-up on #45. Thanks again!

PR #37, opened 2026-07-07
fix(patterns): anchor overload detection to an API Error line, like safeguard

romerjon:
## Problem

The overload phrase patterns are unanchored:

```js
// DEFAULT_OVERLOAD.patterns
'API Error:\\s*(429|500|502|503|504|529)\\b',   // anchored (has "API Error")
'overloaded_error',                              // bare phrase
'temporarily limiting requests',                 // bare phrase
```

`overloadMatch` tests each tail line against them with no context check, so a pane that merely **discusses or quotes** the phrases fires a spurious overload retry. Caught live on my box — the tool's own diagnostic output (a session explaining this very tool) matched `/temporarily limiting requests/`:

```
[INFO] Overload/transient API error detected [matched /temporarily limiting requests/ in:
       "● Key findings: ... "temporarily limiting requests" is a built-in pattern, ..."]
```

`safeguardMatch` already guards against exactly this — it requires an `API Error` line nearby (`SAFEGUARD_ANCHOR`) — but the overload path never got the same discipline.

## Fix

`overloadMatch` now requires an `API Error` line nearby (`OVERLOAD_ANCHOR`, mirroring `SAFEGUARD_ANCHOR`). Every real render satisfies it — the API-429 line and 5xx line carry `API Error` on the same line; a multi-line `overloaded_error` JSON body has `API Error` one line up (within the window). Bare prose mentions don't.

Test-first: prose mentions of `temporarily limiting requests` / `overloaded_error` no longer match; the real API-429 render and a multi-line `overloaded_error` body still do. Self-contained — only `src/patterns.js` + its test. Suite green (243).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,19 @@ All notable changes to this project are documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.5.1] - 2026-06-30
+
+**Upgrade if you installed `0.5.0` from npm.** The `0.5.0` npm artifact was built
+before #29 was merged and shipped without the usage-retry anti-spam fix. `0.5.1`
+includes it. (The git tag `v0.5.0` already contained #29; only the npm tarball was
+behind.)
+
+### Fixed
+- Stop the usage-retry path from spamming an already-resumed session: a lingering
+  limit banner in scrollback no longer re-injects `Continue…` every poll. Detection
+  is now anchored to the live tail, and an `isWorking` gate stops the moment Claude
+  resumes (#29).
+
 ## [0.5.0] - 2026-06-30
 
 This release rolls up everything merged since `0.2.2`, including the API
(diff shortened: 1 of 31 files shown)


cheapestinference (approved):
Verified `088a8a0`. Overload now gates every match on a nearby `API Error` line via `hasNearbyMatch(lines, i, OVERLOAD_ANCHOR)`, mirroring `safeguardMatch` exactly (same symmetric ±6 window). Confirmed against the real exports:

- genuine `API Error: 529 …` and `overloaded_error` / "temporarily limiting requests" bodies sitting under an `API Error:` line → still DETECT.
- bare `overloaded_error` or the 429 phrase in prose, with no `API Error` nearby → no longer false-positives (the point of the PR).
- window boundary is exactly ±6 — distance 6 detects, distance 7 doesn't. No off-by-one.

False-negative risk is low: real overload renders always carry the `API Error:` line, and the body sits within the window. 243/243.

One cross-PR note: this and #34's `c7e39e9` both touch the overload path, but on orthogonal axes — this changes the *match predicate* (the anchor), #34 changes the *windowing* (`contentTail` / `OVERLOAD_MAX_RAW_LINES`). They compose rather than conflict, and neither subsumes the other (#37 alone doesn't fix #34's stale-error-behind-widget case; #34 alone doesn't fix the bare-`overloaded_error` prose case). Just rebase-and-retest whichever lands second. Worth one test for the combined path — an `API Error` line at the `maxRaw` boundary with the body just inside the window — since chrome-stripping/`maxRaw` trimming the anchor while keeping a body line could theoretically drop a real overload.

cheapestinference:
Superseded by #46. I rewrote `master`'s history to clean up commit metadata, which changed the base commits this branch was built on and invalidated the diff here. Recreated as #46 — rebased onto the current `master`, **your commits and authorship preserved**, tests green, review carried over. Nothing needed from you; continue any follow-up on #46. Thanks again!

PR #40, opened 2026-07-09
fix(monitor): leave the wait as soon as the session resumes working (#39)

### LINKED ISSUE #39: Monitor doesn't re-check for a new rate-limit event while in an existing wait state
After detecting a rate-limit banner and entering a wait state, the monitor doesn't re-scan the pane for a newer/different rate-limit event if the session resumes working (e.g. after the user manually types "continue") and then hits a second, separate limit before the original wait timer expires.

Sequence observed:
1. Monitor detects first rate-limit banner, calculates wait time (in my case, miscalculated at ~85065s due to a related stale-timestamp bug — see issue #<link to first issue>).
2. User manually types "continue" to unstick the session (since the calculated wait was wrong).
3. Session resumes working normally for ~3 more hours.
4. Session hits a second, genuine rate limit ("You've hit your session limit · resets 9:40pm (UTC)").
5. Monitor does not detect this second event at all — log shows no new entries after step 1. It's still counting down the original (incorrect) wait timer from step 1, and appears to not re-check pane contents for a new limit message while in that wait state.

Confirmed via ps aux that the monitor process was alive and attached to the correct pane/PID the entire time — this isn't a crashed/dead monitor, it's a monitor that's "parked" waiting on a stale timer and not re-evaluating current pane state in the meantime.

Workaround used: manually kill and restart the monitor process against the same pane/PID to force it back into a clean scanning state:
  kill <old_monitor_pid>
  node .../monitor.js %<pane> <claude_pid> &

Suggested fix: while in a wait state, the monitor should still periodically re-scan the pane for rate-limit patterns (not just count down blindly), so it can detect a new/different limit event that supersedes an earlier miscalculated or stale one — particularly important since this bug compounds directly with the stale-timestamp wait-calculation issue in #38 

Environment: Claude Code v2.1.202, claude-auto-retry via npm, Ubuntu 24.04.4 LTS.

cheapestinference:
Closes #39.

## Problem

In the `waiting` state, `processOneTick` returned `'waiting'` unconditionally until the timer expired:

```js
if (Date.now() < state.waitUntil) return 'waiting';
```

So while a usage wait counts down, the pane is never re-evaluated. If the user manually types `continue` to unstick a wrong/stale wait (e.g. a miscalculated ~24h wait), the session resumes, works for a while, and then hits a **second, genuine limit** — the monitor stays parked on the old timer and never detects it. Confirmed against the code, and the reporter confirmed the monitor process was alive/attached the whole time.

## Fix

Don't take the early return when the pane is working:

```js
if (Date.now() < state.waitUntil && !isWorking(stripped)) return 'waiting';
```

A working pane means the user continued, so we fall through to the **existing** `!isRateLimited || isWorking` gate a few lines down, which already returns to `monitoring` and resets `attempts`. Once in monitoring, the next limit banner re-enters a wait with the correct reset time. Minimal — this reuses the gate rather than adding a separate re-scan path.

## Scope

The trigger is `isWorking`, not re-parsing reset times mid-wait. Between two limits the session necessarily works (that's how you reach a second limit), so this covers the real sequence. Re-parsing a "new" reset time while still rate-limited is deliberately left out — the previous banner lingers in scrollback, so telling a new limit from a stale one isn't reliable without exactly the false-positive risk the tail-anchoring + `isWorking` machinery exists to avoid.

Note: the ~85065s (~23.6h) miscalculated wait the reporter hit is a *separate* bug — the ambiguous-reset-rolls-to-tomorrow issue fixed in #35. This PR is independent: any over-long or stale wait parks the monitor blind, and #35 only makes the wrong-timer case rarer.

## Tests

Added a regression test: with `waitUntil` an hour out and a working pane, the tick now returns `user-continued` and drops to `monitoring` (previously returned `waiting`). Full suite green (219/219).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,19 @@ All notable changes to this project are documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.5.1] - 2026-06-30
+
+**Upgrade if you installed `0.5.0` from npm.** The `0.5.0` npm artifact was built
+before #29 was merged and shipped without the usage-retry anti-spam fix. `0.5.1`
+includes it. (The git tag `v0.5.0` already contained #29; only the npm tarball was
+behind.)
+
+### Fixed
+- Stop the usage-retry path from spamming an already-resumed session: a lingering
+  limit banner in scrollback no longer re-injects `Continue…` every poll. Detection
+  is now anchored to the live tail, and an `isWorking` gate stops the moment Claude
+  resumes (#29).
+
 ## [0.5.0] - 2026-06-30
 
 This release rolls up everything merged since `0.2.2`, including the API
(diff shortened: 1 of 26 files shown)


cheapestinference:
Landed on `master` as `8435940` (fast-forward). GitHub shows this as *Closed* rather than *Merged* because it went in via direct push during the same maintenance pass, but the change is in `master`.

PR #42, opened 2026-07-10
feat(cli): add `reconcile` — re-arm monitors for all live claude panes

cheapestinference:
> ♻️ Recreated from #32 to rebase onto the current `master`. The base branch history was cleaned up, which invalidated the fork-based PR's diff. **All commits are @romerjon's — authorship preserved**; only the base was updated (and merged cleanly, tests green). No action needed from @romerjon; superseded PR #32 will be closed with a pointer here.

---

**Problem.** Monitors are detached processes with no supervising service. If one is killed/crashes — or a claude session is started outside the wrapper — that live session is silently left unmonitored. Only *new* wrapper-launched sessions get a monitor, so coverage degrades over time with no way to restore it short of manual `pgrep`/`kill`/relaunch (error-prone: pane IDs get reused).

**`reconcile`.** Re-arms a monitor for every live tmux pane running claude that isn't already covered, from authoritative tmux + process state:
- maps each live claude to its tmux pane by walking the process tree to the pane's `pane_pid`;
- skips panes that already have a monitor (idempotent — safe to run anytime / on a timer);
- never arms its own pane (`$TMUX_PANE`);
- handles **pane-id reuse** (two claudes resolving to one recycled pane → prefer the foreground one, else highest pid);
- `--dry-run` previews the plan.

The pure parse/plan core is separated from the tmux/ps/spawn runner and unit-tested (11 tests). Uses `spawn` (not `fork`) for the detached monitors so no IPC channel keeps the CLI alive — it exits immediately after arming.

Usage:
```
claude-auto-retry reconcile            # re-arm all uncovered live sessions
claude-auto-retry reconcile --dry-run  # preview
```

Pairs well with a periodic `--user` timer to make monitor coverage self-healing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `/usage-credits` hint) before reading the live tail, so a genuine banner behind a tall
   task widget is still detected (fixes a ~54-min stall) while a banner merely quoted in
   scrollback is not (#34).
+- **`reconcile` / `install-timer` / `exclude-self`** for self-healing monitor coverage: a
+  monitor killed (or a `claude` started outside the wrapper) is re-armed from live tmux +
+  process state, on demand or via a `systemd --user` timer (#32).
 - Safeguard/AUP false-positive auto-retry: when the model's safeguards flag a
   message ("safeguards flagged this message"), re-send a short retry up to
   `safeguard.maxRetries` times, then give up loudly once. Detection is anchored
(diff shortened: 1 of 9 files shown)


cheapestinference:
**🔴 Review verdict: changes requested**

Recreated from #32 (identical code, verified there). **One blocker remains.**

The double-hold the original review was about still reproduces via the **restore** branch of `acquireLock` (`src/reconcile.js`). After `rename(lockPath → grave)` succeeds, `lockPath` is empty during the async `readFile(grave)` + `holderIsLive(...)` window; a concurrent `link(tmp, lockPath)` can legitimately repopulate it, and the restore's `rename(grave, lockPath)` then *atomically clobbers* that fresh holder → two processes believe they hold the lock. Reproduced ~0.3–0.5% under contention (restore-clobbers ≈ double-holds ~1:1).

Fix: make the restore **non-clobbering** — `link(grave, lockPath)` + verify (fails `EEXIST` if a racer repopulated), or re-validate ownership after restoring, rather than an unconditional `rename`.

Nit: the new concurrency tests all call `acquireLock` in one process (shared PID), so they never exercise cross-process interleaving and stay green even with this bug.

cheapestinference:
**🟢 Correction — the double-hold blocker is resolved; withdrawing the changes request.**

My earlier verdict described a double-hold via a *restore branch* (`rename(grave, lockPath)`). That was against the **older** revision of this branch — the lock has since been reworked, and the current `acquireLock` no longer uses rename/steal at all. It now serializes the break through a dedicated **breaker lock**: exactly one run ever removes+recreates a stale main lock, liveness is re-verified *while holding the breaker*, every removal is identity-checked (`releaseLock` re-reads and only unlinks if the id still matches), and creation is atomic (`link()`, EEXIST → back off). That's the correct fix for this class.

I re-verified empirically rather than trusting the read: a mutual-exclusion stress harness (a seeded stale lock + up to 16 concurrent child processes contending to break it, each entering a sentinel-guarded critical section):

- **~8,900 lock acquisitions under contention → 0 double-holds.**

The previous (restore-branch) design reproduced a double-hold at ~0.3–0.5% under the same kind of contention, so ~30–45 would have been expected here. Zero. The unlink→create gap is handled (a fast-path racer that wins the empty path just makes the breaker-holder's `linkCreate` fail EEXIST and loop), and the breaker-cleanup path is identity-checked against re-taking.

No outstanding blockers from me on the lock. 280/280 tests green.

PR #43, opened 2026-07-10
fix(patterns): chrome-aware tail so a limit banner behind a task widget still detects

cheapestinference:
> ♻️ Recreated from #34 to rebase onto the current `master`. The base branch history was cleaned up, which invalidated the fork-based PR's diff. **All commits are @romerjon's — authorship preserved**; only the base was updated (and merged cleanly, tests green). No action needed from @romerjon; superseded PR #34 will be closed with a pointer here.

---

**Symptom (observed live).** A genuine session-limit banner — `You've hit your session limit · resets 2am (Europe/Zurich)` + `/usage-credits to finish…` — sat ~16 lines above the prompt because a tall **todo/task widget**, the `✻ Brewed for 54m` status spinner, the input box, and the footer rendered below it. The tail-scoped `isRateLimited` (last 12 lines, added to fix a scrollback false-positive) never reached the banner, so **no retry fired and the session stalled ~54 min.**

**Why not just widen the window.** That reopens the false positive it was added for: a limit banner buried 15+ lines up *in scrollback* would re-match. Window size alone can't tell the two apart.

**Fix — measure distance in *content*, not raw lines.** The discriminator is what sits *below* the banner: after a live limit only Claude Code UI chrome renders below it (input box, footer, key hints, todo/task widget, status spinner, `/usage-credits` hint, background-agent notices); a stale/quoted banner has real output below it. So the detector strips **trailing chrome** before applying the tail window. A quoted banner with real work below stays out of the window (false-positive still fixed); a live banner behind a tall widget is now reached.

Plus a `/usage-credits` companion backstop scanned over a wider window — a distinctive live-limit string that catches the modern render even if a future widget escapes the chrome allowlist.

Details:
- `isWorking` deliberately keeps a **raw** tail (the `esc to interrupt` / `Retrying` indicators would otherwise be stripped as chrome).
- Same chrome-aware tail applied to the overload detector (same failure class).
- `·`/`•` excluded from the task-item chrome pattern (they also separate `· resets 3pm` in the multi-line render).
- Capture bumped 20→50 for headroom.

+10 tests including the exact reported pane; the scrollback false-positive and the common near-prompt case both still pass. Zero deps.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- **Chrome-aware detection.** Limit/overload/menu detectors now skip trailing UI
+  furniture (input box, footer, key hints, todo/task widget, status spinner,
+  `/usage-credits` hint) before reading the live tail, so a genuine banner behind a tall
+  task widget is still detected (fixes a ~54-min stall) while a banner merely quoted in
+  scrollback is not (#34).
 - Safeguard/AUP false-positive auto-retry: when the model's safeguards flag a
   message ("safeguards flagged this message"), re-send a short retry up to
   `safeguard.maxRetries` times, then give up loudly once. Detection is anchored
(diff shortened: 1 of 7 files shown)


cheapestinference:
**🔴 Review verdict: changes requested**

Recreated from #34 (identical code, verified). The over-broad box rule is genuinely fixed (boxed-input chrome rule anchored to the prompt glyph). **One blocker as a standalone PR.**

This adds `Backgrounded agent` to `CHROME_LINE` (`src/patterns.js`), but the monitoring path (`src/monitor.js`) is **not** gated on `!isWorking`. That gate *and* the `waiting for N background agents` working-pattern live in **#45** (the recreation of #36). So #43 on its own can strip a working background-agent pane's scrollback, pull a stale banner into the tail, and fire `enterUsageWait` on a live session.

Land **#45** together with / before this, or drop the `Backgrounded agent` chrome rule here. #45 effectively supersedes this.

cheapestinference:
Superseded by #45, which is now merged — it carries all of this branch's commits plus the `!isWorking` gate this one needed to avoid the standalone regression. The chrome-aware detection from here is in `master`. Closing; nothing outstanding.

PR #44, opened 2026-07-10
fix(time-parser): grace window so a just-passed reset retries promptly, not in ~24h

cheapestinference:
> ♻️ Recreated from #35 to rebase onto the current `master`. The base branch history was cleaned up, which invalidated the fork-based PR's diff. **All commits are @romerjon's — authorship preserved**; only the base was updated (and merged cleanly, tests green). No action needed from @romerjon; superseded PR #35 will be closed with a pointer here.

---

## Problem

`calculateWaitMs` rolls **any** past reset time a full day forward:

```js
let diff = getTargetTimestamp(parsed.hour, parsed.minute) - now.getTime();
if (diff < 0) diff += 86400_000; // tomorrow
```

That's correct for a reset genuinely hours in the past, but wrong at the reset *boundary*. The monitor can settle on a limit banner minutes-to-an-hour **after** its reset — the chrome-aware anti-spam guard deliberately holds detection until the session stops working, which can push the first detection just past the reset. `diff < 0` then parks the session ~24h even though the limit has effectively cleared.

Observed live: `You've hit your session limit · resets 10am (Europe/Zurich)` detected at 10:03 Zurich → **86273s (~24h)** wait, on a box whose sessions were otherwise fine to continue right after 10am.

## Fix

A 1-hour grace window (`rollPastReset`): a reset within the last hour is treated as just-reset (`diff → 0`, so the wait is only the margin — it retries promptly and self-corrects if still limited); only a reset more than an hour past rolls to tomorrow. Applied to both the exact and ambiguous (am/pm) branches.

```
resets 10am, detected 10:03 Zurich:  60s   (was 86273s)
resets 3pm,  detected 12:48 Zurich:  ~2.2h (unchanged — future reset)
```

## Tests

Test-first: two failing boundary cases (3 min and 55 min past → prompt retry) plus a guard that a reset well over an hour past still rolls to tomorrow. The #6 timezone regressions and all future-reset behavior are unchanged. Full suite green.

Self-contained — touches only `src/time-parser.js` and its test. Surfaced while running #34 on live sessions; kept separate so it doesn't muddy that PR.

### DIFF
--- a/src/time-parser.js
+++ b/src/time-parser.js
@@ -30,6 +30,21 @@ export function parseResetTime(text) {
   return null;
 }
 
+// Reset-boundary grace window. A live limit banner whose parsed reset time is already in
+// the PAST almost always means the reset just happened: the monitor can settle on the
+// banner minutes-to-an-hour after the reset (a session that kept working past it — see the
+// chrome-aware anti-spam guard), and Claude's session limits reset on short cadences, so a
+// past reset time is recent, not "tomorrow". Rolling a just-passed reset a full day forward
+// parks the session ~24h even though the limit has effectively cleared (observed live:
+// "resets 10am" detected at 10:03 → 86273s wait). rollPastReset retries promptly instead
+// (diff→0, so the wait is just the margin); only a reset MORE than the grace window in the
+// past plausibly means the next occurrence is tomorrow.
+const RESET_GRACE_MS = 60 * 60 * 1000; // 1 hour
+function rollPastReset(diffMs) {
+  if (diffMs >= 0) return diffMs;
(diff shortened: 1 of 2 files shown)


cheapestinference:
**🟢 Review verdict: approve**

Recreated from #35 (identical code, verified). **Approve.** The ambiguous-roll blocker is fixed — the else-branch grace-checks the most-recent candidate (`Math.max`) and rolls the earliest occurrence +24h, and it can't grace-fire a genuinely-future reset (that branch is only reached when both candidates are ≤ 0). Boundary sweep monotonic, tests green.

PR #45, opened 2026-07-10
fix: treat "Waiting for N background agents" as working (split from #34)

cheapestinference:
> ♻️ Recreated from #36 to rebase onto the current `master`. The base branch history was cleaned up, which invalidated the fork-based PR's diff. **All commits are @romerjon's — authorship preserved**; only the base was updated (and merged cleanly, tests green). No action needed from @romerjon; superseded PR #36 will be closed with a pointer here.

---

Split out of #34 at your request. **Stacked on #34** (`fix/limit-detection-chrome-aware`) since it builds on that PR's `isWorking`/`WORKING_PATTERNS` — until #34 merges, the diff here also shows #34's commits; the only new commit is the last one.

## The bug (live)

A session hit its session limit, dispatched/awaited a background agent, and kept making progress — but showed none of the streaming/retry footers, so `isWorking` read it as idle. With a stale `resets 3pm` banner lingering above the activity, the waiting branch re-injected `Continue` every grace-window cycle.

## The fix — narrower than my first attempt

You were right that gluing the `Backgrounded agent` **transcript** line into `isWorking` is a worse bug (it lingers → `isWorking` latches true → a genuinely-limited idle session never gets retried). So:

- **Only** the live-only `waiting for N background agents to finish` render goes into `WORKING_PATTERNS`. It vanishes the instant the agent finishes, so it can't pin `isWorking` on an idle limited pane. A finished-agent `Backgrounded agent` transcript is explicitly **not** working — there's a counter-repro test for exactly your scenario (`isWorking` false → the tool still retries).
- The monitoring-path `!isWorking` gate is included **with the justification you asked for**: it mirrors the waiting branch's existing `|| isWorking` guard and closes the re-detection cycle (the `Waiting for N` line alone stops injections, but a stale banner sitting <12 content lines above it still re-enters a wait every tick otherwise). The tradeoff you flagged is real — it widens every working pattern from "skip one injection" to "don't detect while it matches" — so with the `Backgrounded agent` line gone, every remaining working pattern is a **live-only** render by construction, bounding the blast radius. If you'd still rather drop the monitoring-path gate and accept the cosmetic re-cycle, say so and I'll pull it.

Test-first: `Waiting for N` → working; finished-agent transcript → not working and still retried; monitoring tick doesn't enter a wait while a background agent runs. Full suite green (270).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- **Chrome-aware detection.** Limit/overload/menu detectors now skip trailing UI
+  furniture (input box, footer, key hints, todo/task widget, status spinner,
+  `/usage-credits` hint) before reading the live tail, so a genuine banner behind a tall
+  task widget is still detected (fixes a ~54-min stall) while a banner merely quoted in
+  scrollback is not (#34).
 - Safeguard/AUP false-positive auto-retry: when the model's safeguards flag a
   message ("safeguards flagged this message"), re-send a short retry up to
   `safeguard.maxRetries` times, then give up loudly once. Detection is anchored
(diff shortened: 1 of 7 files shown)


cheapestinference:
**🟢 Review verdict: approve**

Recreated from #36 (identical code, verified). **Approve.** The `waiting for N background agents` working-pattern is correct and specific (no false positives against limit/overload/safeguard renders), and the monitoring gate becomes `isRateLimited && !isWorking`, so a live background-agent pane with a stale banner defers instead of injecting.

Note this branch carries #43's commits too (it was stacked on #34), so it **supersedes #43's standalone regression** — landing #45 is what adds the `!isWorking` gate #43 lacks. Merge these together (or merge #45 and close #43).

PR #46, opened 2026-07-10
fix(patterns): anchor overload detection to an API Error line, like safeguard

cheapestinference:
> ♻️ Recreated from #37 to rebase onto the current `master`. The base branch history was cleaned up, which invalidated the fork-based PR's diff. **All commits are @romerjon's — authorship preserved**; only the base was updated (and merged cleanly, tests green). No action needed from @romerjon; superseded PR #37 will be closed with a pointer here.

---

## Problem

The overload phrase patterns are unanchored:

```js
// DEFAULT_OVERLOAD.patterns
'API Error:\\s*(429|500|502|503|504|529)\\b',   // anchored (has "API Error")
'overloaded_error',                              // bare phrase
'temporarily limiting requests',                 // bare phrase
```

`overloadMatch` tests each tail line against them with no context check, so a pane that merely **discusses or quotes** the phrases fires a spurious overload retry. Caught live on my box — the tool's own diagnostic output (a session explaining this very tool) matched `/temporarily limiting requests/`:

```
[INFO] Overload/transient API error detected [matched /temporarily limiting requests/ in:
       "● Key findings: ... "temporarily limiting requests" is a built-in pattern, ..."]
```

`safeguardMatch` already guards against exactly this — it requires an `API Error` line nearby (`SAFEGUARD_ANCHOR`) — but the overload path never got the same discipline.

## Fix

`overloadMatch` now requires an `API Error` line nearby (`OVERLOAD_ANCHOR`, mirroring `SAFEGUARD_ANCHOR`). Every real render satisfies it — the API-429 line and 5xx line carry `API Error` on the same line; a multi-line `overloaded_error` JSON body has `API Error` one line up (within the window). Bare prose mentions don't.

Test-first: prose mentions of `temporarily limiting requests` / `overloaded_error` no longer match; the real API-429 render and a multi-line `overloaded_error` body still do. Self-contained — only `src/patterns.js` + its test. Suite green (243).

### DIFF
--- a/src/patterns.js
+++ b/src/patterns.js
@@ -292,17 +292,28 @@ function toRegexes(patterns) {
   return out;
 }
 
-// Returns { pattern, line } for the first overload pattern matching a tail line, else
-// null. Per-line (not whole-tail) so we can report WHICH line tripped it — invaluable
-// for diagnosing a future false positive (the original bug logged no reason at all).
+// A REAL overload always renders as an `API Error:` line ("API Error: 529 …", "API Error:
+// Server is temporarily limiting requests …", or "API Error: …" one line above a JSON
+// `overloaded_error` body). Requiring that line nearby — the same discipline safeguardMatch
+// uses — keeps the phrase patterns ("temporarily limiting requests", "overloaded_error")
+// from firing when they're merely quoted/discussed in the pane (e.g. a session explaining
+// this tool, or a chat about API errors). Mirrors SAFEGUARD_ANCHOR.
+const OVERLOAD_ANCHOR = [/\bAPI Error\b/i];
+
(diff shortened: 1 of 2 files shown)


cheapestinference:
**🟢 Review verdict: approve**

Recreated from #37 (identical code, verified). **Approve.** Overload detection is now anchored to a nearby `API Error` line, mirroring the safeguard path exactly (symmetric ±6 window); genuine overloads still detect, bare `overloaded_error`/prose no longer false-positives. Composes with #43's overload distance-bound (orthogonal axes — match predicate vs windowing). Rebase-and-retest whichever of #43/#46 lands second.

PR #48, opened 2026-07-10
fix(monitor): keep the overload scraper active once the StopFailure hook is live

romerjon:
## Problem

The event-driven overload path (`StopFailure` markers) only treats `overloaded`/`server_error` as retryable — `rate_limit` (a 429) is deliberately excluded. A **transient API 429** the hook can't emit —

```
API Error: Server is temporarily limiting requests (not your usage limit) · Rate limited
```

is only catchable by the text scraper. But the scraper was gated behind `!state.eventMode`, and `eventMode` **latches permanently** the first time any retryable marker is seen (`monitor.js`, set-once, never reset). So once the hook fired for a pane (e.g. an earlier `server_error`), the scraper was disabled for that monitor's entire life, and a genuinely-stuck 429 was **never retried** — the session had to be resumed by hand.

Observed live: after a `server_error` backoff cycle latched `eventMode`, two panes hit the 429 render above and sat idle at the prompt; the monitor never fired. `events.js` documented this as a "known, accepted gap" — but it strands real sessions, and it's a **regression** from the pre-event-mode monitor, which scraped unconditionally.

## Fix

Drop the `!state.eventMode` gate so the scraper runs on every monitoring tick, event path or not, and remove the now-purposeless `eventMode` field. Safe because:

- The overload patterns are **anchored to an `API Error` line**, so the scraper can't misfire on a session/usage-limit render — that path stays owned by the usage-wait detector (`isRateLimited` is `false` for this 429).
- While a backoff is active (`status === 'overload'`) the tick returns **before** the scraper block, so the event path and scraper never double-detect the same incident.
- The scraper is already `isWorking`-gated and raw-distance-bounded (`OVERLOAD_MAX_RAW_LINES`), so it won't re-fire on a recovered or scrolled-past overload.

## Test-first

A regression test latches the hook via a real `server_error` event, then feeds the exact 429 render with no marker and asserts the scraper still enters overload (was `monitoring`, now `overload-detected`). Removed the test that pinned the old "scraper disabled once `eventMode` latched" behavior. **335/335.**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,8 +57,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   disabled permanently the first time any `overloaded`/`server_error` event latched, so a
   transient API 429 the event path can't emit (`API Error: Server is temporarily limiting
   requests …`) went undetected and the session sat stuck until resumed by hand. The
-  anchored overload patterns can't misfire on a session/usage limit (no `API Error` line),
-  and an active backoff still returns before the scraper, so there's no double-detection.
+  anchored overload patterns can't misfire on a session/usage limit (no `API Error` line).
+  The scraper also skips the exact banner the event path just retried until it clears or
+  changes, so a render lingering after an edge-triggered retry can't open a second backoff
+  (a double injection that would also reset the give-up budget).
 - Monitor no longer stays parked on a stale wait timer once the session resumes:
   while counting down a usage wait, a pane that has resumed working (e.g. the user
   manually typed `continue` to unstick a wrong/stale wait) now drops back to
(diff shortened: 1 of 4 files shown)


cheapestinference:
Reviewed and verified against the branch — the core idea (keep the scraper live so a marker-less API 429 the hook can't emit is still caught) is right and worth having. But removing the `eventMode` latch outright reintroduced a **double-fire**: after the edge-triggered `viaEvent` retry returns to `monitoring` with the overloaded/server_error banner *still on screen* (viaEvent doesn't verify it cleared), the always-on scraper re-detects that same banner and opens a **second** backoff — an extra injection, and `enterOverload`→`resetOverload` zeroes the give-up budget so `maxTotalWaitMinutes` may never trip. Reproduced driving the real `processOneTick` (2 sends/incident vs 1).

I pushed a fix commit (`ba2b6c3`) rather than block: record the banner the event path just retried and have the scraper skip that exact render until it clears or changes. Your 429 goal is untouched (a new/marker-less banner still fires — verified), and I added a regression test for the lingering-same-banner sequence. 347/347 green. Rebased onto current `master` (CHANGELOG conflict resolved). Merging.

PR #49, opened 2026-07-10
fix(reconcile): detect node-launched and launcher-wrapped claude sessions (Finding 6)

romerjon:
## Problem (closes Finding 6 from the #32 review)

`reconcile` identifies sessions by `comm === 'claude'` only. Two real launches show a different comm and were **invisible** — so the self-healing timer never re-armed them once their monitor died:

1. **A claude CLI run under node with `process.title` unset** → comm `node` (the exact shebang case the review flagged).
2. **A session our own launcher wraps in an agent harness that embeds claude** — e.g. `node …/src/launcher.js` → `node …/happier claude` — comm `node`, with the real claude only appearing deeper (or later, during startup). Observed live: a `happier`-launched pane was skipped by `reconcile` during that window.

The old `CLAUDE_COMMANDS = ['claude','node']` constant was dead code; the review asked to "honor the constant or delete it and document the limitation." This honors it — conservatively.

## Fix

Two detectors alongside the existing comm match, shared by `planReconcile` (arming) and `claudePidForPane` (`exclude-self`) via a new `sessionTargetsByPane`:

- **`isNodeClaudeCli`** — a node process whose executed **script** is the claude CLI (basename `claude`, or the `claude-code` package `cli.{js,mjs,cjs}` entry). The script is the discriminator, so `node server.js`, `node build.js claude-prod`, or a bare "claude" as an argument never match.
- **`isLauncher`** — our own `node …/src/launcher.js`, a zero-false-positive marker (it only ever wraps claude). Only fills a pane with **no** direct candidate, and targets the launcher's spawned child (skipping the detached `monitor.js`), matching what the launcher armed at start — so a plain launcher→claude chain isn't double-counted.

Print mode stays skipped (a `claudeArgs` normalizer drops the `node <script>` prefix so `isPrintMode` sees the same shape as a bare `claude …`).

## Test-first

Six `planReconcile` cases: node-launched CLI, `claude-code/cli.js`, a launcher-wrapped agent (targets the child), a plain launcher→claude chain (no double-arm), print mode, and the critical **negative** (`node server.js` / `node build.js claude-prod` must NOT arm). Also verified against **live** process state — all 12 real claude panes detected, zero false positives. **341/341.**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   actual poll interval, not a fixed constant).
 
 ### Fixed
+- `reconcile` now also re-arms claude sessions whose process command isn't `claude`
+  (Finding 6): a claude CLI run under `node` with its process.title unset (shows comm
+  `node`), and a session our own launcher wraps in an agent harness that embeds claude
+  (e.g. `happier claude`) — both were invisible to the `comm === 'claude'` match, so the
+  self-healing timer never re-armed them once their monitor died. Detection stays
+  conservative: only a node process that IS the claude CLI (script basename `claude` or
+  the `claude-code` cli entry) or a pane our `launcher.js` wraps — never a bare node
+  process. `exclude-self` recognizes these sessions too.
 - Monitor no longer stays parked on a stale wait timer once the session resumes:
   while counting down a usage wait, a pane that has resumed working (e.g. the user
   manually typed `continue` to unstick a wrong/stale wait) now drops back to
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -95,8 +95,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - `customPatterns` are matched against the raw last-N lines (unchanged from pre-#34
   semantics), not the chrome-skipped view — the user owns their own tradeoff (#34).
-- Removed the dead `CLAUDE_COMMANDS` constant; documented that reconcile matches
-  `comm === 'claude'` (a bare-`node` session without `process.title` isn't detected) (#32).
+- Removed the dead `CLAUDE_COMMANDS` constant (#32). Reconcile's session detection was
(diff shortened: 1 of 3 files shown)


cheapestinference (dismissed):
Reviewed and verified against the branch (341/341 tests). **Approve** — closes Finding 6.

I ran the real exported `planReconcile` against adversarial process lists. Detection is by *executed script* (`isNodeClaudeCli`: `comm==='node'` + the node script's basename is `claude` / matches `claude-code/.../cli.js`), not a bare `claude` token in args — the right discriminator. Every dangerous class is handled: the reconcile CLI's own node, a detached `monitor.js`, dev-server/editor/npx node processes, and bare `node server.js` all correctly arm nothing; only genuine claude children arm. Self-exclusion holds (the regex requires literal `claude-code/`), so no infinite self-arming.

Non-blocking follow-ups (edge cases): (1) `nodeScript`/`claudeArgs` don't skip *separate-value* node flags, so `node -r <mod> <script>` misreads `<mod>` as the script — a `node -r ... /path/claude` launch is missed (re-opens Finding 6 for preload-instrumented starts) and a contrived `node -r /opt/claude ...` false-matches; skipping known value-taking flags (`-r/--require`, `--import`, `--loader`, `-e`) fixes both. (2) A launcher wrapping print mode (`node .../wrap claude -p`) is armed because `isPrintMode` stops at the wrapper's first positional. (3) Branch-3 trusts the "launcher only spawns claude" invariant without a claude-shape check. None block merge.

cheapestinference:
Heads-up: #51 (macOS/launchd) merged first and reworked the same claude-process detection, so I rebased this onto current `master` and resolved the semantic conflict by **integrating both approaches** rather than picking one — `sessionTargetsByPane` now treats a process as a direct claude via `isClaudeProc` (comm `claude` / macOS full-path / truncated-comm-via-argv0, from #51) **or** `isNodeClaudeCli` (node-launched CLI, from this PR), with print-mode detection using the right arg shape per case. Verified: 346/346 tests (both PRs' suites) plus an adversarial matrix — macOS truncated-comm, node-launched, and plain claude all arm; dev-server / reconcile-self / monitor / print-mode all correctly skip. Merging.

PR #51, opened 2026-07-11
feat(macos): support reconcile + install-timer on Darwin (launchd)

kighfk-aps:
## Problem

The self-healing coverage introduced in #32 is Linux-only in practice. On macOS:

1. **`reconcile` always aborts** with *"pgrep matched processes but printed no monitor args"*: BSD pgrep prints full argument lists with `-l`, not procps' `-a` (BSD `-a` means *include ancestors* and leaves the output PID-only, which `runningFromPgrep` correctly rejects as unverifiable coverage).
2. Even past that, **`planReconcile` sees zero claude sessions**: macOS `ps -o comm=` prints the executable's *full path truncated to 16 chars* (e.g. `/Users/u/.local/`), so the `comm === 'claude'` compare never matches.
3. **`install-timer` hard-requires `systemctl`**, so there is no self-healing timer at all.

## Changes

- **`PGREP_LIST_FLAG`** (`-lf` on Darwin, `-af` elsewhere) for the running-monitor probe, shared with `exclude-self`. The `runningFromPgrep` guard semantics are unchanged.
- **Claude detection falls back to the basename of argv[0] from `args=`** when `comm` doesn't match. A bare `node script.js` session still shows `node` and stays invisible, preserving the documented "install the wrapper for those" behavior on both platforms.
- **`install-timer` installs a launchd LaunchAgent on Darwin** (`~/Library/LaunchAgents/com.claude-auto-retry.reconcile.plist`), `RunAtLoad` + `StartInterval` 300s to mirror `OnStartupSec`/`OnUnitActiveSec`. Two launchd-specific pitfalls are handled in the template with comments:
  - `AbandonProcessGroup` — same reason the systemd unit needs `KillMode=process`: without it the short-lived job's exit reaps the freshly-armed detached monitors.
  - an explicit `PATH` covering both Homebrew prefixes — launchd does not inherit the login shell's PATH, so `spawn tmux` ENOENTs otherwise (reproduced; this made the agent exit 1 on every fire).
  - substituted node/CLI paths are XML-escaped (`renderReconcilePlist`).
- **`ps -o lstart=` (lock start token) pinned to `LC_ALL=C`** so the timer (C locale) and an interactive shell (user locale) always agree on lock-holder identity; Linux keeps taking the `/proc` path and is unaffected.
- `launchd/` added to the npm `files` whitelist; README/CHANGELOG updated.

## Verification

On macOS 15 (arm64), against a live tmux server with 4 claude sessions:

- 340/340 tests pass (new: BSD truncated-comm planning, plist rendering/escaping, template invariants incl. the PATH and `AbandonProcessGroup` keys).
- `reconcile --dry-run` correctly reports all live panes as `already monitored` / `self (excluded)`; with stripped env it reproduced both original failures before the fix.
- `install-timer` → the `RunAtLoad` fire exits 0 (`launchctl print` `last exit code = 0`); `uninstall-timer` boots the agent out and removes the plist; already-running monitors unaffected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **`reconcile` / `install-timer` / `exclude-self`** for self-healing monitor coverage: a
   monitor killed (or a `claude` started outside the wrapper) is re-armed from live tmux +
   process state, on demand or via a `systemd --user` timer (#32).
+- **macOS support for `reconcile` / `install-timer`**: the running-monitor probe now uses
+  `pgrep -lf` on Darwin (BSD pgrep prints full args with `-l`, not procps' `-a`, so
+  reconcile previously always aborted with "cannot verify coverage" on macOS), claude
+  detection falls back to the basename of argv[0] from ps `args=` (macOS `comm=` prints
+  the executable's full path truncated to 16 chars — never "claude" — so the strict
+  compare saw zero claude sessions), and
+  `install-timer` installs a launchd LaunchAgent
+  (`~/Library/LaunchAgents/com.claude-auto-retry.reconcile.plist`, `RunAtLoad` +
+  `StartInterval` 300s, `AbandonProcessGroup` so the freshly-armed detached monitors
+  survive the short-lived job, and an explicit `PATH` covering both Homebrew prefixes —
+  launchd does not inherit the login shell's PATH, so `spawn tmux` would otherwise
+  ENOENT) instead of requiring systemd. The reconcile lock's `ps -o lstart=` start token
+  is now pinned to `LC_ALL=C` so the timer (C locale) and an interactive shell (user
+  locale) always agree on lock-holder identity.
 - Safeguard/AUP false-positive auto-retry: when the model's safeguards flag a
   message ("safeguards flagged this message"), re-send a short retry up to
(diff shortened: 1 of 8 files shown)


cheapestinference (approved):
Reviewed and verified against the branch (340/340 tests, including the new Darwin cases). **Approve.**

No Linux regression: the platform switch (`PGREP_LIST_FLAG`, and the `darwin` guards in `cmdInstallTimer`/`cmdUninstallTimer`) leaves the systemd path byte-for-byte; `processStartToken`'s `LC_ALL=C` only touches the non-Linux `ps` fallback. The launchd plist is well-formed, uses `RunAtLoad + StartInterval 300` with no `KeepAlive` (correct for a periodic reconcile), and `AbandonProcessGroup` is the right analog of the systemd `KillMode=process` so detached monitors survive. The launchctl flow is injection-safe (`execFileSync` arg arrays, XML-escaped plist), and the BSD `pgrep -lf` full-command-line behavior is what `parseRunningMonitors` needs. Solid first contribution.

Non-blocking follow-ups: (1) `uninstallTimerLaunchd` uses a dynamic `import('node:fs/promises')` for `unlink` though the module is already statically imported up top; (2) the `runningFromPgrep` comment still says "macOS/BSD pgrep -af prints PIDs only" (stale after this change); (3) no `StandardOutPath`/`StandardErrorPath`, so macOS reconcile failures are silent (systemd has the journal). None block merge.

PR #56, opened 2026-07-12
fix(monitor): route rate_limit StopFailure markers to the usage-wait path

### LINKED ISSUE #50: Usage-limit rescue is scrape-only and unreliable; the `StopFailure` hook already fires for usage limits (`error: "rate_limit"`) but is only wired to the overload path
## Summary

The usage-limit path (`waiting` state, hours-scale reset) depends entirely on **screen-scraping** the banner out of the pane tail. The session-limit banner is emitted by Claude Code as an inline, one-shot `isApiErrorMessage` turn in response to a submission — it is **not** a persistently-redrawn banner — so whether any 5s poll happens to catch it is a coin-flip. When the poll loses that race, detection never fires, the tool never sends, the banner is never re-emitted, and the session sits idle until the user returns manually.

Meanwhile the **`StopFailure` hook already fires for usage limits** with `error: "rate_limit"` and carries the `session_id` — a reliable, timestamped, non-scraping signal that is currently only used to drive the *overload* (seconds-scale backoff) path, never the *wait-until-reset* path.

**Proposed fix:** treat a `rate_limit` `StopFailure` marker as a first-class usage-limit trigger — read the reset time from that session's transcript (via the `session_id` the marker already carries), enter the existing `waiting` state until the reset, then send the retry. This removes the scrape from the critical path for usage limits entirely.

### Relationship to #38

This overlaps #38 (the fixed `RATE_LIMIT_TAIL_LINES = 12` window) but is more fundamental, not a duplicate. #38 makes the scrape *miss less often* by widening the window. The event-driven path proposed here removes the scrape from the usage-limit critical path entirely, so the banner's on-screen position stops mattering at all — it can't be pushed out of a window that's no longer consulted. In that sense this **supersedes** #38 for usage-limit *detection* when the hook is installed.

#38 still stands on its own, because the scrape doesn't go away completely: it remains for the `/rate-limit-options` menu detection (`isRateLimitOptionsPrompt` / `menuStepsToWaitOption`), overload detection, and as the fallback when the hook isn't installed. So the two are complementary — this issue makes usage-limit detection reliable via the event channel; #38 hardens the scrape that's still needed elsewhere.

## Evidence (two concurrent sessions, same account-wide limit, v0.5.1)

Two panes (`%0`, `%32`) hit the same session limit within ~3 minutes. **Both fired the `StopFailure` hook** — the markers, straight off disk:

```json
// events/_0.json
{ "pane": "%0",  "error": "rate_limit", "session_id": "…", "ts": … }   // written 10:33:57
// events/_32.json
{ "pane": "%32", "error": "rate_limit", "session_id": "…", "ts": … }   // written 10:36:27
```

`error: "rate_limit"` confirms a usage/session limit raises `StopFailure` and passes `isRetryableError()` (events.js). The banner text lives in each session's transcript:

```
You've hit your session limit · resets 12:10pm (Australia/Melbourne)
```

Outcome:
- **`%0` recovered** — but via the *scraping* path getting lucky (log shows `Rate limit detected …` → menu select → wait → `Sent retry message (attempt 1)` → `User already continued`). Its hook marker was **never consumed** (still on disk).
- **`%32` did not recover** — its monitor (confirmed alive throughout) logged nothing but its startup line, sent nothing, and the session sat idle from 10:36 until manual resume ~2h later. Its hook marker was **also never consumed**.

So the hook fired for both panes and rescued neither; the one recovery came from a scrape that happened to win the race. On a prior night the same setup lost the race on the interactive pane and idled for 9 hours.

## The two code-level gaps

1. **Hook markers for usage limits are effectively ignored.** `readEvent` is only reached in the `monitoring` branch (`processOneTick`, monitor.js ~L295–311). A marker written while the monitor is in `waiting` (or outside the `eventMaxAgeSeconds` window) is never consumed and goes stale on disk — which is what happened to both markers above.
2. **Even when consumed, a hook event is routed to the wrong machine.** The `readEvent` branch sets `status = 'overload'` and schedules `nextOverloadWaitMs` (seconds), then relies on a *subsequent scrape* (`isRateLimited`, monitor.js ~L196/L220) to hand off to `enterUsageWait`. It never parses the reset time from the event, so a `rate_limit` event backs off in seconds against a limit that resets in hours.

## Proposed fix (sketch)

- In the `monitoring` (and `waiting`) handling, check for a `rate_limit` `StopFailure` marker **before** the overload routing.
- On such a marker, resolve the transcript path from `session_id` (`~/.claude/projects/<cwd-slug>/<session_id>.jsonl`), read the most recent `isApiErrorMessage` text, and feed it through the existing `parseResetTime` / `calculateWaitMs` → enter `waiting` until the reset, then the existing retry send.
- Keep the scrape as a fallback for cases where the hook isn't installed.

This reuses the entire existing wait-and-retry mechanism; the change is only the *trigger* (reliable event instead of scrape) and the *reset-time source* (transcript instead of on-screen banner).

## Environment

- `claude-auto-retry` v0.5.1 (global npm install)
- Claude Code 2.1.205
- macOS (Apple Silicon), tmux, zsh
- `StopFailure` hook installed via `claude-auto-retry install-hook`

Happy to open a PR (with tests) if the approach looks right.

shaunmower:
Closes #50.

## Problem

A `rate_limit` StopFailure marker was written by the hook and then immediately discarded by the monitor (`isRetryableError` excludes it, so it hit the event-ignored path). That left usage-limit detection entirely dependent on the scraper's tail window — a race: two concurrent panes hitting the same account-wide limit could both get a marker, yet only the one the scraper happened to catch recovered. The other sat idle for hours with an unconsumed marker on disk.

## Fix

- Add a distinct `isUsageLimitError` classifier, kept separate from `RETRYABLE` so the overload (seconds-scale) and usage-wait (hours-scale) routes stay separate.
- On a `rate_limit` marker, resolve the reset time preferring the pane scrape already run this tick (freshest read, and it runs before the event block), falling back to the session transcript the marker's `session_id`/`cwd` point at.
- The transcript lookup (`src/transcript.js`) matches on the `isApiErrorMessage` substring rather than parsing the full record shape, so it survives Claude Code schema drift, and degrades to a no-op (`usage-limit-unresolved`, scraper as safety net) if neither source resolves a message.
- Add `rate_limit` to the hook matcher; the older-hook consume-side guard still stands (`isRetryableError` no longer includes `rate_limit`).

## Tests

17 new tests across events/transcript/monitor/cli, plus an update to the one existing test whose premise this reverses. 369/369 passing.


cheapestinference (changes requested):
Strong PR — I verified the whole pipeline empirically against real Claude Code transcripts on this machine before reviewing, and the core mechanism holds up:

- ✅ A real raw `.jsonl` line (with the actual on-disk encoding — the `·` is stored as a literal UTF-8 char, not a `\u00b7` escape) flows through `findRateLimitMessage` → `parseResetTime` → `calculateWaitMs` and produces the exact right wait (tested with a real `resets 3:30pm (Europe/Berlin)` record: 7,080,000 ms, i.e. reset − now + margin, to the millisecond).
- ✅ `projectSlug` matches real project dirs, including the adjacent-separator case (`/.claude-worktrees/` → `--claude-worktrees-`).
- ✅ A stale `isApiErrorMessage: true` line that is *not* a limit (e.g. a 401 `Please run /login`) does not match the rate-limit patterns, so the worst case degrades to the 5h fallback rather than a bogus parsed wait.
- ✅ The routing/consume semantics and the test coverage for the monitor branch look right, and I'm on board with promoting `rate_limit` to a first-class event — the scrape-only race in #50 is real.

**One blocker, found empirically:** resolving the transcript path from the hook payload's `cwd` fails whenever the session's current cwd differs from its *launch* cwd. The transcript directory slug is derived from the directory the session was **started** in, while the hook's `cwd` is the cwd **at StopFailure time** — and sessions drift (cd, worktrees, subdirs). On this machine, a real rate-limited session resolves to NULL through the PR's path:

- record says `"cwd": "/home/ubuntu/cheapestinference"` → slug `-home-ubuntu-cheapestinference`
- but the transcript actually lives at `projects/-home-ubuntu/<session>.jsonl` (launch dir)

~2% of the 395 sessions on this box diverge in their *first* record already; mid-session drift adds more. The failure is graceful (falls back to the scraper), but it drops the fallback exactly in the long unattended-run scenario this PR exists for.

**Suggested fix — resolve by session id, not cwd:** session ids are UUIDs, unique across projects, so a glob does it robustly:

```js
// try the cwd-derived path first (cheap), then search all project dirs
const matches = await glob(join(configDir, 'projects', '*', `${sessionId}.jsonl`));
```

(or `readdir` over `projects/` and stat for the file — no dep needed). With that, `cwd` becomes an optimization hint instead of a correctness dependency, and the marker's `cwd` field can stay for the fast path.

**Optional hardening, non-blocking:** the returned line is the last `isApiErrorMessage` record of *any* age. Since each record carries a `timestamp`, skipping lines older than a few minutes before the marker's `ts` would rule out parsing a *previous* limit's reset time in the (unlikely) case today's record isn't flushed yet when the monitor reads the marker.

Everything else is mergeable as-is — happy to re-review quickly once the session-glob lands.

cheapestinference (changes requested):
Thanks for this — the routing is exactly what we sketched in #50 (separate `isUsageLimitError`, scrape-first then transcript fallback, graceful degrade), and I verified your ordering claim holds: the pane scrape really does get first shot before the event block. Three things before this can land:

**1. The 08-11 master merge broke the branch.** `test/events.test.js` ended up with two stacked `it('sanitizes the pane id …` openers (lines 76–77) — as an ES module the file no longer parses (`SyntaxError: Unexpected end of input`, reproduced on your head with Node 22), so the suite currently fails. There's also a duplicated comment line above `fileFor` in `src/events.js`. Since v0.7.0 shipped this week you'll need to re-resolve `CHANGELOG.md` too — happily that's the only real conflict left; the code auto-merges clean.

**2. The wait fires, but the retry never does — in exactly the #50 scenario.** I ran your path through a harness: transcript resolves, `enterUsageWait` runs, `status = 'waiting'` — all good. But the waiting branch was built on the assumption that the banner is on screen, and your scenario is defined by it not being there. When `waitUntil` expires, `!isRateLimited(stripped)` returns `'user-continued'` without ever reaching the `sendKeys` — the stranded pane stays stranded. Worse: if stale scrollback contains working-shaped text (`Retrying in 5s … attempt 3/10` from an old deploy log), `resumedAfterLimit`'s no-banner fallback to `isWorking` tears the wait down on the *next tick*, with the marker already consumed. I think you need a `state.viaUsageEvent` flag: while set, skip the no-banner resumed fallback, and at expiry send the retry even without a visible banner. Tests for the expiry tick and the stale-working-text tick would pin both (the PR currently only tests entry).

**3. Two smaller ones.** The hook payload already carries `transcript_path` — persisting that in the marker (cwd-slug as fallback) drops the slug reconstruction and any `CLAUDE_CONFIG_DIR` mismatch between claude's env and the monitor's. And `usage-limit-unresolved` currently consumes the marker before the resolve attempt; the transcript record can flush a beat after the hook fires, so consider leaving the marker in place for a few ticks (the 120s max-age already bounds refire).

With those, this closes #50 properly and I'm glad to merge. The two-pane evidence in the issue remains the best writeup this repo has received.

cheapestinference:
@shaunmower ping — still interested in landing this? The event-driven `StopFailure` rescue is something I actively want (it closes #50), and the review items have been sitting since July. What's pending, concretely:

- transcript resolution breaks when the session's cwd differs from the launch directory (the review has the repro), and
- the tests need to run against real transcript fixtures rather than the synthetic ones.

If you're short on time, say the word (or just don't reply within a week or so) and I'll finish those two items on top of your branch and merge it with your commits and credit intact. If you're still on it, happy to re-review quickly whenever you push.

shaunmower:
Still on it — sorry for the quiet stretch. Just pushed a rebase onto current master plus fixes for both open items:

- transcript_path (the standard hook envelope field) is now preferred over reconstructing the cwd slug — that's the cwd-vs-launch-directory bug.
- Added a real (sanitized) transcript fixture pulled off my own ~/.claude/projects for the transcript-read tests, alongside the synthetic ones.

Also went back through your Aug 13 review and fixed the third item I'd missed: a wait entered off a transcript-resolved marker now tracks that (`viaUsageEvent`), so an absent banner at expiry sends the retry instead of silently giving up as 'user-continued', and stale scrollback can't tear the wait down mid-countdown. Marker consumption on an unresolved read is deferred too, per your note about the transcript flushing a beat late.

550/550 green. Happy to take another look whenever you get a chance.

cheapestinference (changes requested):
Thanks for coming back to this — the rework is the right shape, and I verified it against current `master` (merges clean apart from a CHANGELOG conflict; merged suite 579/0). The `viaUsageEvent` lifecycle and the "don't consume an unresolved marker" change both hold up, and the scrape-first claim is true: the `isRateLimited` scrape at the top of the monitoring branch runs on the same capture before `readEvent`, so the transcript is genuinely the fallback.

Two things before merge, one of them a blocker:

**1. BLOCKER — `transcript_path` never reaches the marker, so the cwd-vs-launch-dir fix is dead code in production.**

`transcript.js` prefers `marker.transcript_path`, but `writeStopFailureEvent` persists only `{pane, error, session_id, cwd, ts}` — the hook payload's `transcript_path` is dropped at write time:

```js
// src/events.js (this branch)
const body = JSON.stringify({ pane, error, session_id: payload?.session_id ?? null, cwd: payload?.cwd ?? null, ts });
```

`readStopFailureEvent` returns that body verbatim, so `resolveUsageLimitLine(ev)` in the monitor always sees `transcript_path: undefined` and falls through to `transcriptPathFor(cwd, session_id)` — the exact path the July review showed resolving `null` when the session's cwd ≠ launch dir. The unit tests pass because `transcript.test.js` constructs the marker object by hand with `transcript_path` set; nothing exercises hook payload → `writeStopFailureEvent` → `readStopFailureEvent` → `readLatestUsageLimitLine`.

Fix is one field plus a round-trip test: `transcript_path: payload?.transcript_path ?? null` in the body (the hook envelope carries it — it's a standard field alongside `session_id`/`cwd`), and an `events.test.js` case mirroring your cwd round-trip that asserts `ev.transcript_path`. I'd also add one end-to-end test that goes through `writeStopFailureEvent` with a payload whose `cwd` is deliberately wrong and `transcript_path` points at the fixture, then `readLatestUsageLimitLine(await readStopFailureEvent(...))` — that's the test that would have caught this.

**2. Not this PR's fault, but it changes what your e2e claim can say today: the fixture's banner is not parseable on `master`.**

Your (real, sanitized) record carries a **weekly** limit with a date — `You've hit your weekly limit · resets Aug 21 at 3pm (Australia/Brisbane)`. Run through current master:

```
findRateLimitMessage(line, [], 12)  → the record line   (ok)
parseResetTime(...)                 → null              → wait = 5.02h  (fallbackWaitHours, not Aug 21 3pm)
isRateLimited(banner-as-pane)       → false             (RESET_PATTERNS require a digit right after "resets")
```

So on master a weekly-limit banner with a date is neither detected by the scraper nor parsed from the transcript — `RESET_TIME_REGEX` and `RESET_PATTERNS` only know `resets 3pm` / `resets at 3:00pm` / `resets in N hours`. That's a real gap your fixture just exposed, and I'll fix it on master separately (date-anchored `resets <Mon> <D> at <time>`), so please don't work around it here. Once it lands, an e2e test through the real fixture asserting `waitUntil` ≈ the record's reset time (and `_waitIsFallback === false`) would pin the whole pipeline — right now such a test would show the 5h fallback.

Minor, non-blocking:
- `usage-limit-unresolved` logs a warning on every tick while the marker is left in place — with a 5s poll and the 120s default `eventMaxAgeSeconds` that's ~24 identical lines per incident. Latch it (log on the first unresolved tick, stay quiet until the marker changes or resolves).
- Rebase onto current master: only CHANGELOG conflicts (both add Unreleased entries), the code merges clean.

Happy to re-review as soon as the marker field is in — with that, this is mergeable from my side.

shaunmower:
Pushed a fix for the blocker plus the minor item from the Aug 26 review:

- `transcript_path` is now actually persisted onto the marker — `writeStopFailureEvent` was dropping it, which made the cwd-vs-launch-dir fix dead code in production. Added a round-trip unit test plus an end-to-end test that goes through `writeStopFailureEvent` → `readStopFailureEvent` → `readLatestUsageLimitLine` with a deliberately wrong `cwd`/`session_id` and the real fixture, so it actually pins the write→read→resolve chain rather than a hand-built marker object.
- The `usage-limit-unresolved` warning is now latched to once per marker instead of once per poll tick.

Left the weekly-limit date parsing gap alone per your note. 553/553 green.

shaunmower:
Merged current `master` in — that pulled in your weekly-limit date-parsing fix too, so the fixture-exposed gap from your last review is now covered as a side effect. Only conflict was CHANGELOG.md (both branches added Unreleased entries); resolved by keeping both. 609/609 green.

Should be clean to re-review now.

Issue #63, opened 2026-07-20
Quoted banner text in a tool-call line is detected as a live rate limit, producing a multi-hour bogus wait


peterellix:
# Quoted banner text in a tool-call line is detected as a live rate limit, producing a multi-hour bogus wait

## Summary

When a pane line contains rate-limit banner text inside a **quoted string** — e.g. a `grep` argument, or this tool's own log output being inspected — `isRateLimited()` returns `true` and `findRateLimitMessage()` returns that line. The reset time is then parsed out of the quoted text, and the monitor parks itself for hours.

The chrome-aware `contentTail` window (added in #34) does not catch this, because the offending line is *recent content*, not stale scrollback. The tail correctly discriminates recent-from-old; this is a different axis — live banner vs. text *about* a banner.

## Reproduction

Verified on `d83d587` (v0.6.0), Node v24.15.0, tmux 3.6b.

```js
import { isRateLimited, findRateLimitMessage } from './src/patterns.js';
import { parseResetTime, calculateWaitMs } from './src/time-parser.js';

const TAIL = 12; // RATE_LIMIT_TAIL_LINES, monitor.js:14

const pane = [
  '● Bash(grep -c "5-hour limit reached - resets 3pm (UTC)" ~/.claude-auto-retry/logs/2026-07-18.log)',
  '  ⎿  3',
  '╭──────────────────────────╮',
  '│ >                        │',
  '╰──────────────────────────╯',
  '  ⏵⏵ auto mode on',
  '~/project | opus 4.8 | 47k/200k | v2.1.214',
].join('\n');

isRateLimited(pane, [], TAIL);        // => true
findRateLimitMessage(pane, []);       // => the Bash(...) line
calculateWaitMs(parseResetTime(findRateLimitMessage(pane, [])), 60, 5, new Date('2026-07-18T20:30:00Z'));
                                      // => 66,720,000 ms  (18.52 hours)
```

**Control** — replacing only the quoted argument with an unquoted placeholder yields `isRateLimited(...) === false`, confirming the quoted text is what triggers it.

## Observed vs expected

| | |
|---|---|
| Observed | `true`; monitor enters `waiting` with an 18.52h `waitUntil` |
| Expected | `false`; a shell command that merely *mentions* a limit is not a live limit |

## Real-world impact

This is not hypothetical. On 2026-07-18 a monitor matched a `grep` argument and parked for **22.52 hours**:

```
[...]

PR #66, opened 2026-07-25
feat(windows): add native Windows support

yonikremer:
## Windows Support: Task Scheduler Timers, win32 Process Enumeration, and Robust Spawning/Parsing

  ### Overview

  This PR implements full Windows compatibility for claude-auto-retry, including:

  • Windows Task Scheduler Integration: Wires install-timer to Windows Task Scheduler via PowerShell, enabling self-healing monitor coverage repeating every 5 minutes at logon.
  • Process List & Monitor Gathering: Replaces Unix ps/pgrep with PowerShell-based Get-CimInstance Win32_Process queries on Windows to extract PID, parent PID, command lines, and check monitor/launcher statuses.
  • Quote-Aware Tokenizer: Adds a robust splitArgs utility to correctly parse command-line arguments containing spaces (e.g. quoted folder and node executable paths), which is common in default Windows node installations.
  • Safe Batch Script Spawning: Spawns Windows .cmd/.bat binaries natively using shell: true rather than custom cmd.exe /c wrappers, closing path-with-spaces execution errors.

  ### Key Changes

  • README.md: Document Windows setup instructions, limitations, and shell integrations.
  • cli.js: Integrate installTimerWindows and buildWindowsTaskEncodedCommand using base64-encoded UTF-16LE payloads for scheduled tasks.
  • launcher.js: Detect Windows batch files and route execution safely via shell: true.
  • reconcile.js: Implement windowsProcessList, processStartToken (using PowerShell timestamps), and a race-free holderIsLive check.
  • test: Expand the test suite to cover Windows Task Scheduler arguments generation, WSL paths routing, and process start token stability.

  ### Verification

  • Checked locally on Windows 11.
  • All 385 unit tests passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -419,7 +419,7 @@ claude-auto-retry uninstall-hook [dir]  # Remove it
 claude-auto-retry reconcile        # Re-arm a monitor for every live claude pane not covered
 claude-auto-retry reconcile --dry-run   # Preview without arming
 claude-auto-retry install-timer    # Run reconcile every 5 min (systemd --user on Linux,
-                                   # launchd LaunchAgent on macOS)
+                                   # launchd LaunchAgent on macOS, Task Scheduler on Windows)
 claude-auto-retry uninstall-timer  # Remove the timer
 claude-auto-retry exclude-self     # Keep THIS session unmonitored (durable, self-expiring)
 ```
@@ -466,7 +466,9 @@ sessions get a monitor. Two commands restore and maintain full coverage:
   step. On Linux this is a `systemd --user` timer (enable `loginctl enable-linger $USER`
   once if you want it to run while logged out); on macOS it is a launchd LaunchAgent in
   `~/Library/LaunchAgents` (LaunchAgents only run while you are logged in, which is fine —
-  the tmux server it reconciles lives in your login session too).
+  the tmux server it reconciles lives in your login session too); on Windows it is a Task
+  Scheduler task that runs shortly after logon and repeats every 5 minutes while you are
+  logged in.
 
 **Excluding a session.** To keep a specific session *unmonitored* (e.g. one where you're
 pasting rate-limit text and don't want any auto-retry), run `claude-auto-retry
@@ -490,6 +492,16 @@ pruned, since staleness can't be detected). Prefer the PID form; you can also ha
 | macOS | `brew` | Fully supported |
 | Arch Linux | `pacman` | Fully supported |
 | Alpine | `apk` | Fully supported |
+| Windows 10/11 | Manual install (Git Bash / MSYS2 / WSL) | Print mode works natively; interactive sessions need tmux |
+
+### Windows notes
+
+- The test suite passes on native Windows 11.
+- `claude -p` / `--print` works without tmux.
+- Interactive sessions require tmux. Install tmux through your POSIX environment (Git Bash, MSYS2, or WSL) and run `claude-auto-retry install` from that shell.
+- `reconcile` uses PowerShell (`Get-CimInstance Win32_Process`) to enumerate processes.
+- `install-timer` uses Windows Task Scheduler. It requires tmux to be available on the
+  task's PATH, so it prepends the detected Git Bash / MSYS2 `usr\bin` directories.
 
 ### Requirements
 
@@ -612,7 +624,7 @@ node --test --watch test/             # Watch mode
 
 - **New rate limit patterns** — If you see a Claude Code rate limit message that isn't detected, open an issue with the exact text.
 - **Fish shell support** — Auto-install for fish shell (currently manual).
-- **Windows support** — WSL works, but native Windows would need a different approach.
+- **Windows shell integration** — Auto-install for PowerShell / cmd profiles (currently use Git Bash / MSYS2 / WSL).
 - **Notification integration** — Desktop/Slack notification when rate limit detected or when Claude resumes.
 
 ## Related Projects
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -8,7 +8,7 @@ import { execFileSync, spawn } from 'node:child_process';
 import { fileURLToPath } from 'node:url';
 import { writeStopFailureEvent, isRetryableError } from '../src/events.js';
 import { sweepStaleStatus } from '../src/status-file.js';
-import { reconcile, excludeSelf, parseRunningMonitors, PGREP_LIST_FLAG } from '../src/reconcile.js';
+import { reconcile, excludeSelf, parseRunningMonitors, PGREP_LIST_FLAG, findRunningMonitors } from '../src/reconcile.js';
 
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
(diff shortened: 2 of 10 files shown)


cheapestinference (changes requested):
This is a serious, well-put-together PR — the Task Scheduler integration via `Register-ScheduledTask` with an `-EncodedCommand` payload (and tests that decode and assert on it), the `Win32_Process` CIM process list normalized to the Unix `ps` shape, the CRLF normalization, and the `holderIsLive` re-read to close the slow-`processStartToken` race are all good work, and the README is honest that interactive sessions still need tmux. Test suite passes here too (385/385 on Linux). Three things need to change before merge, though — the first one is a hard blocker:

**1. The `patterns.js` changes must come out of this PR (verified false positive).**

Adding `/budget exceeded/i` to *both* `LIMIT_PATTERNS` and `RESET_PATTERNS` makes it self-pairing: a single line satisfies the limit+reset proximity check, which is the core false-positive defense of the whole detector. Verified against this branch:

```js
isRateLimited('> The deploy failed because the AWS budget exceeded the quota we set last month.')  // → true
isRateLimited('what does "budget exceeded" mean in the vercel dashboard?')                          // → true
```

Any pane where the *conversation* mentions those words now looks rate-limited, and the monitor will start a bogus wait — exactly the failure class we just shipped fixes for in #63 (a quoted banner caused a 22-hour phantom wait). `/next cycle\b/i` has the same problem in milder form, plus no parseable reset time — and for a *billing-cycle* budget message, waiting is arguably the wrong rescue anyway (the quota comes back next month, not next hour).

Please split the Kimi/on-prem patterns into their own PR so Windows support isn't held up. Note the detector already supports user-supplied `customPatterns` via config — for on-prem gateway messages that may be the right home rather than shipping them to everyone. If they do go in core, they need to keep the reset-anchor discipline (e.g. match the actual companion line `will be refreshed in the next cycle` as RESET, and require a real LIMIT line nearby — never self-pair).

**2. `shell: true` with an args array is unsafe on Windows.**

In `launchInteractive`/`launchPrintMode`, `spawn(cmd, cmdArgs, { shell: true })` makes Node join the args into a shell command *without escaping* (this is what DEP0190 warns about). A print-mode prompt containing shell metacharacters — `claude -p "check foo & bar > baz"` — gets interpreted by `cmd.exe`, corrupting the command or worse. Safer options: resolve the `.cmd` shim to the underlying JS entry and spawn `node` directly, or build the full command string yourself with proper `cmd.exe` quoting. Whatever route, add a test with a metacharacter-laden prompt.

**3. Gate the quote-aware `splitArgs` to the Windows process list.**

On Unix, `ps` argv never contains quoting — but prompts often contain apostrophes, and the new tokenizer changes how they parse (`claude -p what's up` → `what`, `s`, `up` where it used to be `what's`). That silently shifts the inputs to `isPrintMode` / `looksLikeClaude` / `claudeArgs` on platforms that work today. Windows `CommandLine` genuinely needs quote-awareness; Unix should keep the whitespace split. Easiest: do the quote-stripping/tokenizing inside `windowsProcessList` normalization, keep the shared helpers as they were.

Small nits, non-blocking:
- `findRunningMonitors` pgrep pattern: `'node .*src/monitor\.js'` — in a JS string `\.` collapses to `.`, so the regex is now `monitor.js` (any-char dot). The original had `\\.`.
- Duplicated comment line above `paneForPid` ("Walk pid → ppid chain…" twice).
- PR is from your fork's `master` — fine for now, but a topic branch makes follow-up pushes safer.

With the patterns hunk dropped and 2–3 addressed this is very mergeable — the Windows plumbing itself looks solid, and I'd like to land it.

cheapestinference (changes requested):
Hi @yonikremer — thanks for this, and sorry for the slow review. There's real demand for Windows and the two core design calls here are exactly right: `Register-ScheduledTask` with an `-EncodedCommand` payload sidesteps both elevation and quoting hell, and `Get-CimInstance Win32_Process` is the right replacement for `ps`/`pgrep`. The test coverage on the encoded-command builder (apostrophes, ampersands) is appreciated.

That said, I can't merge it in its current shape, for three concrete reasons:

1. **It's now based on a launcher that no longer exists.** v0.7.0 (released this week) rewrote the launch path: env crosses via a 0600 snapshot file, clean exits reap the tmux session, and `new-session` retries transient server teardown. Your branch conflicts textually in `launcher.js`/`reconcile.js`, and the win32 `gather()` branch drops `#{socket_path}` from the pane listing, which would reintroduce the multi-server monitor-masking bug on Windows. A rebase here really means a re-test on Windows, since the interactive path you verified has changed underneath.

2. **`shell: true` for `.cmd` shims is unsafe.** Node doesn't quote `args` (or the command) when `shell: true`, so a prompt like `claude -p "explain this & that"` gets re-split and `&` is interpreted by cmd.exe — prompt text becomes shell syntax — and a `claude.cmd` under a path with spaces still breaks. This needs explicit `cmd.exe /c` with proper quoting (or resolving the shim to its underlying script). Two smaller ones in the same vein: a PowerShell failure in the win32 `processStartToken` returns `null`, which `holderIsLive` reads as "stale" and steals a live holder's lock (null should mean "can't verify → assume live"), and `windowsProcessList`'s args-normalization mishandles unquoted exe paths with spaces — worth extracting into a pure function with unit tests, since it feeds all the classification logic.

3. **Scope**: the `budget exceeded` / `next cycle` pattern additions are a separate feature with a real design question (`budget exceeded` as its own reset pattern means retrying forever on the fallback wait against a quota that doesn't reset on a clock) — could you split those into their own PR/issue? Same for the portable goodies (CRLF normalization, the `holderIsLive` re-read race fix, the `findRunningMonitors` dedupe) — I'd merge those quickly on their own.

One logistical ask: could the next iteration come from a feature branch rather than your fork's `master`? Every fork sync rewrites the PR, which makes review rounds harder for both of us. Happy to help with the rebase details — and the env-snapshot semantics on Windows (POSIX file modes are mostly a no-op there, so the snapshot rides on `%USERPROFILE%` ACLs) deserve a comment in the code as part of it.

cheapestinference:
@yonikremer ping — is Windows support still on your radar? The foundation here is solid and I'd like to land it, but the open review items are all Windows-behavior ones I can't verify myself (no Windows environment on my side):

- the self-pairing "budget exceeded" pattern (a verified false positive in the review),
- the `shell: true` spawn injection surface, and
- `splitArgs` drifting from Unix semantics.

That last part matters: since I can't test on Windows, this one can only merge with an author who can run it there. If you're still up for it, push whenever and I'll re-review promptly. If life got in the way — no problem, just say so; I'd then leave the PR open tagged as needing a Windows-equipped author rather than close it.

yonikremer:
I still want it as a feature. But I have exams and won't have the time to fix that:( If you/someone else want to edit anf fix I would be happy. For now, I am using my unmerged fork.

thnhmai06:
How about #79?

cheapestinference:
@yonikremer no worries — good luck with the exams, and thanks for getting it this far. @thnhmai06 yes, #79 is the better framing for native Windows (a backend abstraction rather than per-call special-casing), and I've replied there with what it would take.

Plan for this PR: I'll leave it open for now as the reference implementation — the Task-Scheduler timer, CRLF normalization and the process-list work are reusable by whoever picks #79 up. If nobody has by late September I'll close it with credit rather than let it drift, and #79 stays as the tracking issue. If you get time back before then and want to finish the three review items, just say so and I'll re-review promptly.

Issue #58, opened 2026-07-13
tmux new-session fails on Windows: "invalid -e: invalid environment variable name"


DevZero-AIKI:
On Windows (Git Bash / MSYS2, tmux 3.3.6 via MSYS2 package), every `claude` invocation through the shell wrapper fails immediately with:

```
psmux: invalid -e: invalid environment variable name
[claude-auto-retry] Failed to create tmux session: Command failed: tmux new-session -d -s claude-retry-... -e ACLOCAL_PATH=... -e ... -e ProgramFiles(x86)=C:\Program Files (x86) -e ...
```

**Root cause**

In `src/launcher.js`, `createTmuxSession()` forwards every entry of `process.env` to `tmux -e`:

```js
const envArgs = [];
for (const [k, v] of Object.entries(process.env)) {
  if (k.startsWith('TMUX')) continue;
  if (v == null) continue;
  envArgs.push('-e', `${k}=${v}`);
}
```

Windows exposes several standard environment variables whose names are not valid POSIX identifiers — e.g. `ProgramFiles(x86)` and `CommonProgramFiles(x86)` contain parentheses. `tmux -e` (like `export`/`putenv`) requires names matching `[A-Za-z_][A-Za-z0-9_]*`, so tmux rejects the whole `new-session` call, and *no* claude-auto-retry session is created — the tool is effectively non-functional out of the box on any Windows + Git Bash setup.

**Suggested fix**

Filter env var names to the POSIX-valid subset before building `envArgs`:

```js
for (const [k, v] of Object.entries(process.env)) {
  if (k.startsWith('TMUX')) continue;
  if (v == null) continue;
  if (!/^[A-Za-z_][A-Za-z0-9_]*$/.test(k)) continue; // skip names invalid for tmux -e (e.g. Windows "ProgramFiles(x86)")
  envArgs.push('-e', `${k}=${v}`);
}
```

**Environment**

- OS: Windows 11
- Shell: Git Bash (MSYS2)
- tmux: 3.3.6 (installed via MSYS2/pacman)
- claude-auto-retry: freshly installed via `npm i -g claude-auto-retry` + `claude-auto-retry install`

Verified locally: applying the one-line filter above resolves the crash and `tmux new-session` succeeds.

---

Willst du, dass ich es stattdessen ins Deutsche übersetze, oder soll ich versuchen herauszufinden, woran das Öffnen der Datei bei dir scheitert (z. B. welches Programm du dafür nutzt)?

[...]

PR #67, opened 2026-07-25
fix: whitelist tmux launch environment

quantumnic:
## Summary
- Stop exporting the full shell environment into tmux-launched Claude sessions
- Keep only the small set of variables the launcher actually needs
- Reduce accidental local secret exposure by excluding unrelated env vars from the tmux session

## Changes
- Introduced a safe tmux session environment whitelist
- Added coverage to ensure unrelated secrets such as `GITHUB_TOKEN` and `AWS_SECRET_ACCESS_KEY` are not forwarded

## Verification
- `npm test` → 380 passed, 0 failed

### DIFF
--- a/src/launcher.js
+++ b/src/launcher.js
@@ -38,6 +38,24 @@ function shellEscape(s) {
   return "'" + s.replace(/'/g, "'\\''") + "'";
 }
 
+// Only propagate the minimum environment needed for a healthy Claude/tmux launch.
+// This avoids copying every shell secret into a long-lived tmux session while still
+// preserving the auth/config vars the launcher actually depends on.
+export const TMUX_SESSION_ENV_VARS = [
+  'PATH', 'HOME', 'USER', 'SHELL', 'TERM', 'LANG', 'LC_ALL', 'LC_CTYPE',
+  'CLAUDE_AUTO_RETRY_LAUNCH_WRAPPER', 'CLAUDE_CONFIG_DIR',
+  'ANTHROPIC_API_KEY', 'OPENAI_API_KEY', 'HTTP_PROXY', 'HTTPS_PROXY',
+  'NO_PROXY', 'NODE_OPTIONS', 'NVM_DIR', 'NODE_PATH',
+];
+
+export function buildTmuxSessionEnv(env = process.env) {
+  const out = {};
+  for (const key of TMUX_SESSION_ENV_VARS) {
+    if (env[key] != null && env[key] !== '') out[key] = env[key];
+  }
+  return out;
+}
+
 // After the tmux session's own claude-auto-retry process exits, the pane falls through
 // to a plain shell so the user isn't dropped straight out of tmux. Use the user's actual
 // login shell (env.SHELL) rather than a hardcoded `bash` — tmux's own `default-shell`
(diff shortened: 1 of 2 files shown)


cheapestinference (changes requested):
Thanks — the motivation is sound (a long-lived tmux session shouldn't inherit every shell secret, and `tmux show-environment` makes them casually visible), and unifying the ≥3.0 and <3.0 paths behind one `buildTmuxSessionEnv` is a nice cleanup. But as written the whitelist breaks real, supported Claude Code setups, and it does so silently.

**1. It strips variables Claude Code needs for auth/routing.**

- **Bedrock users**: `CLAUDE_CODE_USE_BEDROCK` + `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` / `AWS_SESSION_TOKEN` / `AWS_REGION` / `AWS_PROFILE`. The new test explicitly asserts `AWS_SECRET_ACCESS_KEY` is dropped — for a Bedrock user that's their *auth credential*, not an unrelated secret. Their session just fails with a confusing auth error.
- **Vertex users**: `CLAUDE_CODE_USE_VERTEX`, `ANTHROPIC_VERTEX_PROJECT_ID`, `CLOUD_ML_REGION`, `GOOGLE_APPLICATION_CREDENTIALS`.
- **Proxy/gateway users**: `ANTHROPIC_BASE_URL`, `ANTHROPIC_AUTH_TOKEN`, `ANTHROPIC_MODEL`, and the lowercase `http_proxy`/`https_proxy`/`no_proxy` variants (plenty of tooling only reads lowercase).
- **Everything else Claude Code and the tools it spawns rely on**: any other `CLAUDE_CODE_*`/`ANTHROPIC_*` flag, `SSH_AUTH_SOCK` (git push inside the session dies without it), `GPG_TTY`, `XDG_*`, `TMPDIR`, `EDITOR`/`VISUAL`, `GIT_*`, `COLORTERM`/`TERM_PROGRAM`.

The failure mode is the worst kind: nothing errors at launch — the user's session quietly can't auth or can't push, with no hint the launcher filtered their environment.

**2. Suggested shape.** A static name list can't keep up with what users legitimately need. Two options I'd accept:

- *Prefix allowlist + exact names*: keep your exact list, plus prefixes `ANTHROPIC_`, `CLAUDE_`, `AWS_`, `GOOGLE_`, `XDG_`, `LC_`, `GIT_`, `SSH_`, lowercase proxy vars — and an escape hatch like `CLAUDE_AUTO_RETRY_ENV_EXTRA=FOO,BAR` for anything else.
- *Invert it*: keep forwarding everything, but filter out names that are invalid for `tmux -e` (`/^[A-Za-z_][A-Za-z0-9_]*$/`) and optionally a small explicit denylist. Filtering invalid names would also fix #58 (Windows chokes on names like `ProgramFiles(x86)`), which makes this branch attractive — worth coordinating.

**3. Threat-model caveat worth a comment in the code**: the tmux session runs as the same user, so anything that can read the session env could read `~/.aws` anyway — the win here is hygiene (accidental `tmux show-environment` exposure, env dumps in logs), which argues for erring on the side of *not* breaking auth setups.

Also heads-up: #66 touches `src/launcher.js` too, so whichever lands second will need a rebase.

Happy to re-review quickly once the allowlist covers the Bedrock/Vertex/proxy/agent cases (with the test flipped to assert they *survive*).

cheapestinference:
Thanks again for this, @quantumnic — you were right about the underlying problem, and you flagged it before we had a proper fix. As of v0.7.0 (released today) the mechanism this PR patches no longer exists: the environment no longer crosses into tmux via `-e` argv pairs or inline exports at all. It's written to a `0600` JSON snapshot in a `0700` dir, only the file *path* touches the command line, and the inner launcher consumes and unlinks it (see `writeEnvSnapshot`/`consumeEnvSnapshot` in `src/launcher.js` and the `[0.7.0]` Security notes in the CHANGELOG). That closes the `/proc/<pid>/cmdline` and `tmux show-environment` exposure completely, while keeping full env fidelity — which the earlier review round showed we need, since a static whitelist silently breaks Bedrock/Vertex/gateway auth setups.

So I'm closing this as superseded rather than unmerged on merit — the concern was real and it's now fixed at the root. Would be glad to see more from you on the tracker.

Issue #61, opened 2026-07-15
The claude auto retry doesn't seem to work at all for me on the mac


pinkfish:
I installed the auto-retry from npm, I tried it with just scraping but not finding anything.  I then installed the hooks and this also doesn't seem to be triggering anything.  When I run the status and logs I get this:

No activity today

Then when I try the logs they are completely empty.

The hook is there in the zsh file and I can see the tmux running and I can also see the node script running in the background.  It is just not find the messages and restarting.

Not quite sure what is wrong since everything seems to be working, could it be because macos uses zsh?

an8s:
Same for me today

pinkfish:
I think it might be because tmux is using bash instead of zsh.  I just changed the default shell to zsh and it looks like it might be working now.  The logs find the session anyway.

Put:
set-option -g default-shell /bin/zsh

into the ~/.tmux.conf

pinkfish:
No, looks like this is not helping...

cheapestinference:
Thanks for the report. Everything you describe (tmux up, node running, hook in your zshrc) suggests the monitor is alive but the pane scrape isn't matching — so let's narrow it down.

Could you share:

1. Version: `claude-auto-retry --version` and `tmux -V`, plus your Claude Code version.
2. When a limit banner is actually on screen, the last ~30 lines of the pane: `tmux capture-pane -p -S -30` (feel free to redact paths). I want to see the exact banner text your Claude Code renders — the wording varies by version and the patterns key off it.
3. The status file and log paths: `ls -la ~/.claude-auto-retry/` and the contents of the newest file in `logs/`.
4. Confirmation that the pane the monitor attached to is the same one running claude — if you launched claude in one pane and the monitor is watching another, it'll sit idle with "No activity today".

zsh itself isn't the problem (the hook path is shell-agnostic), so this is almost certainly either a pane-mismatch or a banner-wording mismatch. The capture-pane output when a limit is showing is the most useful thing here.

pinkfish:
I did finally get it to work.  I ran zsh inside of tmux and then ran claude
and this seemed to work.  Also when I stopped trying to run two claude/tmux
sessions.

daften:
Doesn't work for me, but I have a different message I don't see in the table. I'm using bash anyway and the hook is there in .bashrc.

Message I get is "You've hit your monthly spend limit." It seems like it's a monthly limit thing, but it's just because the extra usage is above budget, it's still the 5h limit being hit.

[...]

PR #70, opened 2026-08-08
fix(monitor): re-read the reset time while waiting, shortening a fallback wait

nyxaria:
## The bug

Observed live: a pane hit its session limit and the monitor sat in `waiting` with `attempts: 0` for ~4 hours *after* the limit had already reset.

```
status file : {"status":"waiting","waitUntil":1786224431,"attempts":0}
              waitUntil = 22:27:11 BST
banner      : "You've hit your session limit · resets 6:20pm (Europe/London)"
```

Not a detection failure — detection was fine the whole time. Running the shipped detectors against that exact pane capture:

```
isRateLimited(tail=12): true
findRateLimitMessage  : "⎿  You've hit your session limit · resets 6:20pm (Europe/London)"
parseResetTime        : {"hour":18,"minute":20,"timezone":"Europe/London","ambiguous":false}
calculateWaitMs       : 60000 ms
```

The chain:

1. The `/rate-limit-options` menu appeared at 17:26. The monitor handled it correctly — navigated to "Stop and wait for limit to reset", pressed Enter, logged `menu-confirmed`.
2. It then parsed the reset time **from that menu screen**. The menu does not always render a reset line, so `parseResetTime` returned `null` and `calculateWaitMs` took the fallback branch: `5h * 3600 + 60` = **18060s** exactly — which is what the log line said.
3. The `waiting` branch returns early on every tick while `Date.now() < waitUntil`, and never re-reads the pane. So the banner Claude Code prints *right after* confirming — which does carry the time — was ignored for the entire fallback.

A second pane took the same path 21 seconds earlier and got the same 18060s; it only recovered because the user continued it by hand.

## The fix

The `waiting` branch now re-derives the wake-up from the live banner each tick and pulls it earlier when the standing wait is too long.

Two bounds, both load-bearing:

- **Shorten only.** An unreadable screen re-derives the same multi-hour fallback, and a banner drifting through the tail must never be able to postpone the retry.
- **Only before the first send of an episode** (`attempts === 0`). After a send, `waitUntil` is the 30s cooldown; after max-retries it is the long give-up backoff. Both are deliberately unrelated to the reset time, and re-deriving against an already-passed reset collapses them to the margin — burning every attempt in a few ticks.

Also extracts `usageWaitUntil()` so first detection, the menu path and the correction share one derivation rather than a third copy.

## Verification

Replaying the real stuck state (the actual `waitUntil` value, against the actual pane capture) through the fixed monitor:

```
before : waitUntil = 08/08/2026, 22:27:11
tick 1 : wait-corrected   waitUntil = 18:45:02 (60s away)   sent: []
tick 2 : waiting                                            sent: []
tick 3 : retried          attempts = 1                      sent: ["Continue where you left off…"]
```

Corrects once, does not re-fire, does not double-send.

Five new tests in `test/monitor.test.js`: shortening to an already-passed reset, shortening to a still-future reset, never lengthening, no correction after a send, and falling straight through to the retry when the corrected wait has already elapsed.

Full suite on this branch: **421 pass, 0 fail**.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   server that can be days stale, so a rotated `ANTHROPIC_API_KEY` or fresh proxy var
   quietly never reached `claude` with zero diagnostic. The degrade stays; the silence
   goes — a stderr warning now names the cause.
+- **A fallback wait is now corrected once the real reset time appears on screen.** The
+  `/rate-limit-options` menu does not always render a reset line, so confirming "Stop and
+  wait" could commit the `fallbackWaitHours` default (5h) — and the waiting branch returned
+  early on every tick and never looked at the pane again, so the banner Claude Code prints
+  immediately after confirming, which *does* carry the time, was ignored for the whole
+  fallback. Observed live: a session whose limit reset at 18:20 sat parked until 22:27 with
+  `attempts: 0`, ~4 idle hours, while the banner naming 18:20 was on screen the entire time.
+  A wait derived from an unreadable screen is now latched as a fallback and re-derived from
+  the live banner each tick until a real reset time is found; waits that already came from a
+  real reset time are never re-parsed. Confirming the menu starts a fresh retry episode, so
+  the correction still applies when the menu re-renders after a retry has been sent.
 
 ## [0.7.0] - 2026-08-13
 
--- a/src/monitor.js
+++ b/src/monitor.js
@@ -78,14 +78,53 @@ async function checkForeground(tmuxAdapter, pane, config) {
   return { ok: false, fg, isShell };
 }
 
-function enterUsageWait(state, stripped, config) {
+// Reset text on screen → the absolute instant to wake up at. One source of truth for the
+// three callers that need it: first detection, the /rate-limit-options menu path, and the
+// mid-wait correction below. `parsed` is surfaced so callers can tell a real reset time
+// from the fallbackWaitHours default that calculateWaitMs returns for an unreadable screen.
+function usageWaitUntil(stripped, config) {
   const message = findRateLimitMessage(stripped, config.customPatterns);
-  state.lastRateLimitMessage = message;
   const parsed = message ? parseResetTime(message) : null;
-  state.waitUntil = Date.now() + calculateWaitMs(parsed, config.marginSeconds, config.fallbackWaitHours);
+  const until = Date.now() + calculateWaitMs(parsed, config.marginSeconds, config.fallbackWaitHours);
+  return { message, parsed, until };
+}
+
+function enterUsageWait(state, stripped, config) {
+  const { message, until } = usageWaitUntil(stripped, config);
+  state.lastRateLimitMessage = message;
+  state.waitUntil = until;
   state.status = 'waiting';
   state._gaveUp = false;
   return 'waiting';
+}
+
+// Re-derive the wake-up from the LIVE banner while already waiting, and pull it earlier
+// when the standing wait is too long.
+//
+// A wait computed from a screen that carried no parseable reset time lands on the
+// fallbackWaitHours default — potentially hours past the real reset. The
+// /rate-limit-options menu is the common source: it renders the options but not always the
+// reset line, while the banner Claude Code prints right after confirming DOES carry the
+// time. The waiting branch returned early on every tick and never looked at the pane again,
+// so that banner was ignored for the entire fallback (observed live: "resets 6:20pm"
+// detected via the menu at 17:26, monitor parked until 22:27 with attempts:0 while the
+// banner sat on screen the whole time — ~4h of a reset session doing nothing).
+//
+// Two bounds, both load-bearing:
+//   - SHORTEN ONLY. An unreadable screen re-derives the same multi-hour fallback, and a
+//     banner drifting through the tail must never be able to postpone the retry.
+//   - Only before the first send of this episode (attempts === 0). After a send, waitUntil
+//     is the 30s cooldown — and after max-retries it is the long give-up backoff. Both are
+//     deliberately unrelated to the reset time, and re-deriving against an already-passed
+//     reset collapses them to the margin, burning every attempt in a few ticks.
+function correctUsageWait(state, stripped, config) {
+  if (state.attempts > 0) return false;
+  if (!isRateLimited(stripped, config.customPatterns, RATE_LIMIT_TAIL_LINES)) return false;
+  const { message, parsed, until } = usageWaitUntil(stripped, config);
+  if (!parsed || until >= state.waitUntil) return false;
+  state.waitUntil = until;
+  state.lastRateLimitMessage = message;
+  return true;
 }
 
 function enterOverload(state, overload, rand) {
@@ -147,13 +186,11 @@ export async function processOneTick(state, tmuxAdapter, pane, config, isAlive, 
       await new Promise(r => setTimeout(r, 80));
     }
     await tmuxAdapter.sendKey(pane, 'Enter');
-    // Parse the reset time straight from the menu text, so the wait does not depend
-    // on the limit banner still being visible afterward.
-    const message = findRateLimitMessage(stripped, config.customPatterns);
-    state.lastRateLimitMessage = message;
-    const parsed = message ? parseResetTime(message) : null;
-    state.waitUntil = Date.now() + calculateWaitMs(parsed, config.marginSeconds, config.fallbackWaitHours);
-    state.status = 'waiting';
+    // Parse the reset time straight from the menu text, so the wait does not depend on the
+    // limit banner still being visible afterward. The menu does not always RENDER a reset
+    // line, though — that lands on the fallbackWaitHours default, which correctUsageWait
+    // then pulls back in once Claude Code prints the real banner post-confirm.
+    enterUsageWait(state, stripped, config);
     state._menuCooldownUntil = Date.now() + cooldown;
     return 'menu-confirmed';
   }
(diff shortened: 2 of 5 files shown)


cheapestinference (changes requested):
The diagnosis is exactly right, the shorten-only + first-send-only shape is the correct instinct, and the PR description's forensics (replaying the real stuck state) are the standard I wish every PR met. I want this fix in. But an adversarial pass — executing `processOneTick` against constructed pane states, not just reading — found real holes in the guards as implemented, two of which recreate the parked-idle bug this PR fixes. Requesting changes on the four below; the rest are take-or-leave.

## Blocking

**1. The gate and the parse read different windows, so non-banner text can collapse a *correct* wait.** `correctUsageWait` gates liveness on `isRateLimited` over the 12-line *content tail*, but `usageWaitUntil` → `findRateLimitMessage` parses the bottom-most `RESET_PATTERNS` line of the **full ~120-line capture** — no tail bound. Verified live against the branch: a pane with a genuine "resets in 5h" banner plus an unmasked lower line like `The API said to try again in 2 minutes before the limit window rolls` (model prose, user-typed text, wrapped tool output past the mask's documented wrap gap) returns `wait-corrected` and the correct 5h wait collapses to ~180s. The monitor then wakes into the still-live limit, burns `maxRetries` at 30s spacing, and lands in the give-up hold *before the real reset* — the exact failure being fixed, now on waits that used to be correct. And unlike master, which ran this parse once at detection, the PR re-runs it every tick, so *any* bogus line drifting through during a multi-hour wait gets a shot, and shorten-only means the earliest bogus time wins.

**2. Root-cause fix for (1), and it makes the whole thing cheaper: latch fallback-ness instead of re-deriving.** `usageWaitUntil` surfaces `parsed` precisely so callers can tell a real reset from the fallback — but `enterUsageWait` throws it away. Store it (`state._waitIsFallback = !parsed`) and gate `correctUsageWait` on it. Both fallback entries (menu path and unreadable first detection) enter with `parsed === null`, so everything the PR claims still works — while a wait derived from a *real* reset time becomes invulnerable to mid-wait re-parsing, which also stops ~600 dead re-parses (2 stripAnsi+mask passes + up to ~7 `Intl.DateTimeFormat` constructions per tick) over a normal 5h wait. Keep the tail-bounded parse from (1) as well: fallback waits still re-parse, and they should read the same window the gate reads.

**3. The `attempts === 0` gate defeats the fix on the menu-after-send path, and the refactor flips `_gaveUp` semantics.** Verified live: waiting expires → retry sent (`attempts = 1`) → still-limited session re-renders the `/rate-limit-options` menu → menu branch confirms → menu has no reset line (the PR's own premise) → `enterUsageWait` commits the 5h fallback **with the correction permanently blocked** (`attempts > 0`). The post-confirm banner with the real time one minute away is ignored for 5 hours — the ~4h-idle bug, surviving on this flow. Related semantic delta in the same refactor: the old inline menu block set only message/waitUntil/status, but `enterUsageWait` also clears `_gaveUp`, so a maxed-out monitor whose menu re-renders now reports a healthy countdown (`gaveUp: false`) that on expiry immediately re-gives-up without sending. Both point the same direction: a *confirmed* menu is a fresh episode — reset `attempts` alongside `_gaveUp` in that path (or key the correction on episode state rather than the raw counter). Please pin the menu-after-send flow with a test either way.

**4. The new tests are wall-clock dependent.** `bannerAt()` derives the banner from the host clock with no timezone: at host time 00:00–00:29, `bannerAt(-30min)` renders *yesterday's* `resets 11:45pm`, which parses ~23.5h into the future → `waiting` instead of `wait-corrected`/`retried` — any CI or dev run in that half-hour window is deterministically red. (Plus a fall-back-transition-day flake on the `+2h` case via late-side resolution.) Pin the clock / inject `now` / use an explicit timezone the way `test/time-parser.test.js` does.

## Non-blocking, worth taking while in there

- `correctUsageWait` sets `state.lastRateLimitMessage` on ticks that fall through to `retried`/`user-continued`; neither handler clears it, so the *next* plain `waiting` tick logs a spurious `Rate limit detected: … Waiting 29s...` (with the send cooldown as the "wait"). Pairing set-with-clear was an invariant before this PR.
- The shorten-only comparison has no epsilon and `until` isn't idempotent across ticks (two `Date.now()`/`new Date()` reads per derivation), so a persistent parseable banner can re-"correct" by milliseconds and re-log through a long wait. `until <= state.waitUntil - 1000` quiets it.
- `FALLBACK_MS` in the tests hand-copies the fallback formula; `calculateWaitMs(null, margin, fallbackHours)` is exported and returns exactly that value — call it so the tests can't drift from the real path.
- The `wait-corrected` log block is the third clone of the secs+log+clear pattern in `startMonitor`, and the incident narrative now lives in three places (comment, CHANGELOG, test preamble — already drifting: "10:27pm" vs "22:27"). One helper + one canonical narrative location.

Also: please rebase — master just picked up `2716ca0`, which makes `findRateLimitMessage` skip statusline/chrome rows (#61). It's complementary to this PR (a ccusage usage-meter row was hijacking the parse the same way finding 1 describes, just deterministically), and the correction path should be tested on top of it.

The underlying incident this PR pins is real and worth fixing — with the fallback latch + episode-reset in place I think this lands quickly. Happy to re-review fast.

nyxaria:
Thanks so much for the review @cheapestinference 

All four blocking findings addressed, plus the four non-blocking ones. Rebased onto `79b9a19`.

I reproduced both parked-idle findings against the branch before changing anything, driving `processOneTick` with constructed panes. Numbers below are `waitUntil` deltas, same script run against three trees:

| scenario | upstream master | PR as reviewed | now |
|---|---|---|---|
| **1b** correct 5h wait, then `…try again in 2 minutes…` drifts into the pane mid-wait | 301min (never corrects) | 301min → **3min** | 301min → 301min |
| **3** retry sent, menu re-renders, post-confirm banner 1min away | 301min (never corrects) | 301min → 301min | 301min → **1min** |

## What changed

**2 (root cause of 1) — latch fallback-ness.** `enterUsageWait` records `_waitIsFallback = !parsed`; only a fallback wait is correctable. Cleared on the three waits that are deliberately unrelated to a reset time (post-send cooldown, max-retries hold, `user-continued`) and on a successful correction, so the correction fires exactly once and a wait derived from a real banner is never re-parsed.

**1 — tail-bounded parse.** `findRateLimitMessage` takes `tailLines` and reads the same chrome-aware window `isRateLimited` gates on; mask computed full-pane and sliced, as in the detectors. `0` keeps the full scan for print mode.

One correction to the framing, since it changes what the tail bound is worth: **the tail bound does not fix your example on its own.** Your prose line sits *below* the banner, so it's inside the 12-line window either way and the bottom-up scan still prefers it. What actually saves the correct wait in 1b is the latch. The bound's real value is the consistency you asked for — a fallback re-parse can't reach a stale banner *above* the window that the gate never saw.

That leaves a case neither change covers, which I want to flag rather than quietly fold in: when the prose is already on screen **at first detection**, the bottom-up preference parses it instead of the banner and commits a 3min wait. Measured identical on true upstream master (3min), so it's pre-existing and orthogonal to this PR — but it's the same shape as #61, just with prose instead of a statusline row. Happy to take it as a follow-up; it seemed wrong to widen this PR's scope into changing detection-time parse order.

**3 — a confirmed menu is a fresh episode.** `enterUsageWait(…, { fresh: true })` clears `attempts` alongside `_gaveUp` on the menu path, which fixes both halves you identified: the correction is no longer blocked after a send, and the `_gaveUp` clear is now semantically earned rather than an accident of the refactor. Pinned by `corrects the menu fallback even after a retry has already been sent`, and by `a maxed-out episode still gives up rather than looping on the same banner` for the give-up path.

**4 — deterministic test clock.** Banners now render in the fixed-offset zone where the current instant reads as ~midday and carry that zone explicitly, so no relative offset can cross a date boundary and no DST transition applies (`Etc/GMT` zones have none). Your 00:00–00:29 case is gone along with the fall-back-day flake, without threading a clock through `processOneTick`.

**Non-blocking, all taken.** `lastRateLimitMessage` is set only on the branch that logs it (with a test pinning that a correction falling through to `retried` leaves it null); 1s epsilon on the shorten comparison; tests derive the fallback from `calculateWaitMs(null, …)`; the secs+log+clear block is one `logWait` helper; and the incident narrative now lives in the CHANGELOG only, with the code comment and test preamble pointing at it.

## Tests

443 pass. The one failure is `reconcile.test.js` "grants mutual exclusion across real concurrent processes" — pre-existing timing flake, fails 3/3 on clean master on my box (it races real processes and this machine has ~20 live monitors); unrelated to these files.

cheapestinference (approved):
All four blocking findings and all four non-blocking ones addressed — and verified against the branch, not just read:

- **Latch coverage**: `_waitIsFallback` is set at every `enterUsageWait` entry (including the overload→usage transition) and cleared on the three deliberately-unrelated waits (post-send cooldown, give-up hold, `user-continued`) plus on successful correction. Checked every mutation site.
- **Window consistency**: `usageWaitUntil` and `correctUsageWait` read the same chrome-aware `contentTailRange` window `isRateLimited` gates on, mask computed full-pane and sliced. `--print` mode keeps the unbounded scan via the `0` default.
- **Fresh episode on menu confirm**: `fresh: true` clears `attempts` and `_gaveUp` together, and the menu-after-send flow is pinned by a test, as is the give-up hold staying uncorrectable.
- **Deterministic clock**: the `Etc/GMT±N` midday-zone construction kills both the 00:00–00:29 window and the DST-day flake without threading a clock through `processOneTick`. Nice solution — simpler than injecting `now`.
- Suite: **445 pass, 0 fail on my box** — including `reconcile.test.js`'s mutual-exclusion test, so that flake didn't reproduce here.

Also appreciated the honest correction on what the tail bound does and doesn't carry in finding 1 — you're right that the latch is what saves that example, and right to keep the detection-time parse-order case (parseable prose already on screen at first detection) out of scope. Please do file it as a follow-up issue; it's the same shape as #61.

One mechanical thing before merge: the release re-cut this afternoon moved the env-snapshot CHANGELOG entry from `[Unreleased]` into `[0.7.0]` (v0.7.0 shipped today), which is exactly where your entry anchors — so the PR now shows a CHANGELOG-only conflict. A trivial rebase onto current master fixes it (your entry goes under the now-empty `[Unreleased]`); happy to resolve it at merge time instead if you prefer. Approving now — this lands as soon as that's sorted.

Issue #71, opened 2026-08-13
Spend-limit banner ("You've hit your org's monthly spend limit", no reset time) is never detected


cheapestinference:
## The render

Team/org accounts (and individual accounts whose extra-usage budget is exhausted) get a limit banner that talks about the **spend budget**, not the session window, and carries **no reset time**:

```
You've hit your org's monthly spend limit · run /usage-credits to raise it, or visit claude.ai/admin-settings/usage
  ⎿  You've hit your org's monthly spend limit · run /usage-credits to raise it, or visit claude.ai/admin-settings/usage
     /usage-credits to finish what you’re working on.
```

(Variant on individual plans: `You've hit your monthly spend limit.`)

Reported by @zzrough and @daften in #61. Replaying the captures through the shipped detectors: **not detected**, for two independent reasons:

1. **No reset anchor.** `isRateLimited` deliberately requires a reset-time line near the limit line — that anchoring is what keeps conversation *about* limits from triggering bogus retries — and this render has none.
2. **The `org's` apostrophe defeats the limit-pattern shape.** `hit your … limit` is matched as `your` + up to three `\w`-word tokens + `limit`; `org's` isn't a `\w` token, so the org variant doesn't even register as a limit line.

## Why it's worth handling (carefully)

Both reporters confirmed that **waiting works**: the underlying 5h session block resets and the session comes back with plenty of quota, even though the banner talks about the monthly budget. Overnight runs on Team plans currently strand until morning on a limit that cleared hours earlier.

The reason this wasn't blindly enabled before: if the monthly budget is *genuinely* exhausted, no amount of waiting helps, and a retry loop into a dead session is worse than doing nothing. Any fix must stay bounded.

## Plan

- Detect the spend-limit render **only** in the live region and **only** with its `/usage-credits` companion nearby (the companion is a distinctive UI string, already used as the live-limit backstop) — same false-positive discipline as the main path.
- Fix the pattern shape so `org's` / possessives don't break the limit match.
- Route it into the existing usage-wait as a limit with **unknown reset** → the bounded `fallbackWaitHours` wait, standard `maxRetries` cap.
- Once #70 lands, a mid-wait correction picks up the real `resets <time>` banner if/when the session block resets and shortens the wait to the true instant — so the fallback is a ceiling, not the wake-up time.
- If the wait expires and the pane still shows the spend banner with no reset time, that's the genuinely-exhausted case: give up per the normal max-retries/give-up path rather than looping.

[...]

PR #74, opened 2026-08-14
fix(patterns): prefer a live banner over reset-shaped prose below it (#73)

### LINKED ISSUE #73: findRateLimitMessage prefers reset-shaped prose over the live banner at detection time
Follow-up from the review of #70, filed at @cheapestinference's request. Same shape as #61 — a non-banner line winning the bottom-up scan — but with ordinary conversation text rather than a statusline row, so the chrome allowlist can't classify it away.

## Behaviour

`findRateLimitMessage` returns the **lowest** line matching `RESET_PATTERNS`. `/try again in \d+\s*(hours?|minutes?|h|m)/` is one of those patterns, and it matches ordinary prose. Anything reset-shaped rendered *below* a live banner therefore outranks it.

Measured against current `master` (`6f3da65`, i.e. with #70's latch and tail bound in place):

```
model prose:
  banner        : "You've hit your session limit · resets 5:20pm (Etc/GMT+5)"
  parsed instead: "⏺ The API said to try again in 2 minutes before the limit window rolls"
  -> waiting, wait = 3min (expected ~300min)

user-typed text:
  banner        : "You've hit your session limit · resets 5:20pm (Etc/GMT+5)"
  parsed instead: "❯ it told me to try again in 2 minutes, is that right?"
  -> waiting, wait = 3min (expected ~300min)
```

Repro — three lines is enough, no scrollback depth needed:

```js
const pane = [
  "You've hit your session limit · resets 5:20pm (Etc/GMT+5)",   // 5h out
  '',
  '⏺ The API said to try again in 2 minutes before the limit window rolls',
].join('\n');
const s = createMonitorState();
await processOneTick(s, mockTmux(pane), '%0', DEFAULT_CONFIG, () => true);
// s.waitUntil is ~3 min out, not ~5 h
```

## Why it matters

The monitor wakes ~3 minutes in, finds the limit still live, and spends `maxRetries` sends at 30s spacing — landing in the give-up hold with hours of the real limit window left. From there nothing is sent until the banner clears, so the session sits idle exactly as in #70's incident.

Note #70's `_waitIsFallback` latch does **not** cover this: the wait is derived from a `parsed` result, so it is treated as authoritative and never revisited. That's correct behaviour for the latch — the input to the parse is what's wrong, not the decision to trust it.

## Why the existing guards don't catch it

- **Chrome allowlist (#61)** — deliberately anchored to rendered furniture. Model output and user-typed text are content by definition; widening the allowlist to reach them is exactly the loose-glyph matching the `CHROME_LINE` comment warns against.
- **Tool-echo mask (#63)** — only masks `● Name(` blocks and their children. A plain assistant line or a user prompt line has no header to latch.
- **Tail bound (#70)** — the prose sits *below* the banner, so it's inside the window either way; bottom-up still prefers it.

## Possible directions

No strong opinion on which, and each has a cost worth weighing:

1. **Prefer a candidate that also looks like a banner** — require a `LIMIT_PATTERNS` match on the same line (or within the existing `WINDOW`) before accepting a `RESET_PATTERNS` line, falling back to today's behaviour if none qualifies. Cheap; would have picked the banner in both cases above. Risk: multi-line TUI renders where limit and reset are genuinely on separate lines already rely on the loose form, so the fallback has to stay.
2. **Reuse the pairing the gate already found.** `isRateLimited` locates a limit line and a nearby reset line; that pair is strictly better evidence than an independent bottom-up scan, and threading it through would make the gate and the parse agree by construction rather than by convention.
3. **Tighten `/try again in …/`** so it needs an error-ish anchor. Narrowest change, but it weakens a pattern that exists precisely to catch terse API wording.

Happy to send a PR for whichever you prefer — (2) looks like the real fix to me, since it removes the class rather than another instance of it, but it's a bigger change to `findRateLimitMessage`'s contract than (1).

nyxaria:
Closes #73.

## Heads-up: this is not option (2)

You picked (2) and sketched a tiered `findRateLimitMessage`. I built it, and three of its properties broke on an adversarial pass — all traceable to **qualifying a candidate by looking at a neighbour**. Implementing your sketch literally (strong-LIMIT = `LIMIT_PATTERNS` minus `/try again in/i`; tier 1 same-line; tier 2 reset with a strong-LIMIT within `WINDOW`; tier 3 unchanged):

- **tier 2 reopens #73 one render up.** With prose below the multi-line banner (`⚠ You've hit your limit` / `· resets 3pm (UTC)`), the banner's own limit line is what qualifies the prose — tier 2 returns `⏺ The API said to try again in 2 minutes …`, which is the reported bug.
- **tier 1 inverts freshness.** On `[banner resets 11:30am, output, "Please try again in 15 minutes"]` the one-line banner satisfies both vocabularies, so tier 1 returns the stale 11:30am: 1 min when the reset had just passed, 21h when it hadn't. And since that parses, #70's latch marks it non-correctable.
- **prose becomes the qualifier.** Strong-LIMIT still contains bare `/usage limit/i` and `/rate limit/i`, and #73's premise is that conversation about the limit is on screen, so `❯ why do I keep hitting the usage limit?` qualifies `⏺ it said to try again in 2 minutes` two lines down.

*(An earlier revision of this description also listed a doubled `candidate→anchor→reset` window and a tail-boundary fallback. Those were defects of my own neighbourhood implementation, not of your sketch — tier 2 as you wrote it returns the reset line itself, one hop, and tier 3 keeps master's answer when qualification fails. Retracted.)*

So I dropped the neighbourhood entirely. If you'd still rather have the tiered shape, say so and I'll implement it with these cases pinned — but I think per-line eligibility is the better trade.

## What this does instead

**A reset-shaped line is eligible unless its shape makes it conversation.** The scan is otherwise untouched: still bottom-up, both historical fallbacks still there, so a vetoed line degrades to exactly what the scan returned before rather than to null. Three per-line signals, no neighbour lookups:

1. **The prompt glyph.** `❯`/`>` introduce the user's own input, which is never a render.
2. **Sentence punctuation.** A render does not end in `.`/`?`/`!`; a sentence does — including one whose *wrapped* continuation happens to begin with the clause (`  try again in 2 minutes, so I'll wait.`). `capture-pane` is called without `-J` (`src/tmux.js`), so a wrap arrives as its own line, and a rule keyed on what *leads* a line reads that continuation as a render.
3. **What follows the reset clause.** A render stops at the time, give or take a timezone qualifier or the rest of a compound duration — `· resets 3pm (UTC)`, `resets 3am NY` (the #6 fixture), `resets in 3 hours 15 minutes`. A sentence keeps going: `…resets 9am tomorrow according to the header`.

The message bullets `⏺`/`●` are deliberately **not** a signal on their own — #63's fixture has a live banner rendering as `● You've hit your session limit · resets 2:10am`, so vetoing the glyph demotes a real render. They mark a line only when what follows them is *not* a limit or reset clause, i.e. the bullet introduces a sentence about one. That test is an exemption inside the veto, so it can only ever rescue a line, never demote one.

### Why a veto rather than an allowlist of banner shapes

The first commit did it the other way round — a line was eligible if it *led* with a known limit or reset clause. Re-reading its own claim of "never worse than master" turned up two defects, fixed in the second commit:

- **Freshness inverted for unrecognised renders.** An allowlist demotes every render this file doesn't know, so an eligible *stale* banner outranks a live one below it. `Claude usage limit reached. Resets at 2pm` — already in the suite as an `isRateLimited` fixture — buries the phrase behind a word, so under a stale `resets 11:30am` it lost the scan: master commits a **16min** wait, the allowlist committed **22h**, latched non-correctable. That is the same failure I cite above for rejecting the tiered proposal, reintroduced from the other side.
- **Wrapped prose still won.** The wrap can land right before the clause, leaving `  try again in 2 minutes, so I will wait` leading with a reset. The exact bug in #73, one wrap away; #73's own repro line is 71 columns wide.

A veto only ever demotes what it can identify, so an unrecognised render keeps the freshness ordering it had before. `LIMIT_PATTERNS`, `RESET_PATTERNS` and `SPEND_LIMIT` stay byte-identical to master — nothing asks "is there a limit nearby" any more, so your invariant is satisfied by there being no such qualification at all.

### Renders the veto wrongly claimed, found in review (third commit)

Five shapes the pre-#73 scan returned and the second commit demoted. All five now agree with master again, each pinned by a test:

| render | why it was vetoed |
|---|---|
| `● You've hit your session limit · resets 2:10am` | the bullet glyph, before it became conditional |
| `· resets in 3 hours 15 minutes` | the clause matcher stops at the first unit, so `hours 15 minutes` read as a sentence carrying on |
| `Please try again in 4 hours and 12 minutes` | same |
| `5-hour limit resets 3pm, weekly limit resets 9am Monday` | two clauses of the *same* matcher: `exec()` without `/g` measured from the first |
| `· resets 3pm - see above`, `· resets 3pm / 10:30 UTC` | a lone separator counted as one of the two tail words |

## Verification

Differential over a 792-pane corpus (11 banner forms × 12 noise shapes × stale banner × chrome), scored on "does the result carry the live banner's reset time":

```
this branch       792/792 correct
pre-#73 master     72/792      -> 720 fixed, 0 regressions
first cut (f77e878) 390/792     -> 402 fixed, 0 regressions
```

`isRateLimited`, `resumedAfterLimit`, `isRateLimitOptionsPrompt`, `isWorking`, `menuStepsToWaitOption`: **0 behavioural diffs** vs pre-#73 master across 24,565 calls (4,913 panes × 5 detectors).

Monitor level:

| scenario | master | this PR |
|---|---|---|
| banner + model prose below | 3min | **300min** |
| banner + user-typed text below | 3min | **300min** |
| banner + *wrapped* prose below | 3min | **300min** |
| multi-line render + prose below | 3min | **150min** |
| stale banner + fresh unrecognised render | 16min | **16min** (first cut: 1320min) |
| banner alone (control) | 300min | **300min** |

**Mutation testing**, since the first draft's tests passed while the code was wrong — 13 mutants, all killed, with one test per signal shaped so the other two don't fire:

```
veto removed                        duration tail not consumed
prompt-glyph signal dropped         parentheses not stripped
bullet signal dropped               separator tokens count as words
bullet exemption removed            tail budget 0 / infinite
sentence-punctuation dropped        clause scan not global
eligibility pass scans top-down     skip() dropped from the eligibility pass
```

Suite: **491 pass, 0 fail**. (`reconcile.test.js`'s cross-process hold test is flaky on master too — ~50% locally, unrelated to this change.)

## Known boundary

An **unglyphed, unpunctuated continuation that begins with the clause and ends within two words** still wins — `│  try again in 20 minutes  │` (a wrapped input-box row), or a wrap whose sentence carries on into the *next* line:

```
⏺ The API told me I should
  try again in 2 minutes
  before the window rolls
```

This one is not closable per line: a *banner* that wraps has exactly the same shape — `⚠ You've hit your session limit ·` / `  resets 8:40pm (Europe/London)` — and vetoing it would drop the only line carrying the time. Both cases are unchanged from master, which returns the same lines.

## Commits

1. `f77e878` — the fix itself: prefer a live banner over reset-shaped prose below it.
2. `98ebc46` — state the rule as a veto rather than an allowlist, closing the freshness inversion and the wrapped-prose hole.
3. `8e911ff` — narrow the veto to shapes that are only ever conversation, restoring the five renders above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,16 +13,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   …` matches ordinary English — so a model sentence ("The API said to try again in 2
   minutes …") or the user's own question, rendered *below* the banner it discusses, stole
   the parse and turned a 5-hour wait into ~3 minutes. The monitor then woke into the
-  still-live limit and burned `maxRetries` before the real reset. The discriminator is
-  position within the line, not vocabulary: Claude Code *renders* a banner's limit and
-  reset clauses as the leading content of a line (behind at most the `⎿`/`└`/`⚠`/`·` echo
-  glyphs), while prose carries the same words mid-sentence — the discipline the
-  spend-limit banner already uses, generalised. Lines that merely mention a limit are now
-  skipped over, so the banner above them wins. Eligibility is decided per line and never
-  by inspecting a neighbour, which keeps the bottom-up scan — and therefore freshness —
-  exactly as it was: a stale banner still loses to any newer one below it. Anything the
-  scan does not recognise as a render falls through to the previous behaviour rather than
-  to no match at all.
+  still-live limit and burned `maxRetries` before the real reset. The discriminator is the
+  shape of the line, not its vocabulary, and it is stated as a *veto*: a reset-shaped line
+  stays eligible unless it carries a message/prompt glyph (`⏺`/`●`/`❯`) or a sentence
+  running on past the reset clause ("…try again in 2 minutes, so I'll wait"), which is what
+  catches prose whose wrapped continuation happens to begin with the clause. Lines that
+  merely mention a limit are vetoed, so the banner above them wins. Eligibility is decided
+  per line and never by inspecting a neighbour, which keeps the bottom-up scan — and
+  therefore freshness — as it was: a stale banner still loses to any newer render below it,
+  including renders this file does not recognise. Anything vetoed falls through to the
+  previous behaviour rather than to no match at all.
 - **The org/monthly spend-limit banner is now detected (#71).** Team/org accounts (and
   individual accounts whose extra-usage budget is exhausted) get "You've hit your org's
   monthly spend limit · run /usage-credits …" — undetected for two independent reasons:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,15 +15,32 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   the parse and turned a 5-hour wait into ~3 minutes. The monitor then woke into the
   still-live limit and burned `maxRetries` before the real reset. The discriminator is the
   shape of the line, not its vocabulary, and it is stated as a *veto*: a reset-shaped line
-  stays eligible unless it is the user's input row (`❯`/`>`), ends in sentence punctuation,
-  runs on past the reset clause ("…resets 9am tomorrow according to the header"), or sits on
-  a message bullet introducing something other than a limit or reset clause. The last two
-  are what catch prose whose *wrapped* continuation happens to begin with the clause. Lines
-  that merely mention a limit are vetoed, so the banner above them wins. Eligibility is decided
-  per line and never by inspecting a neighbour, which keeps the bottom-up scan — and
-  therefore freshness — as it was: a stale banner still loses to any newer render below it,
-  including renders this file does not recognise. Anything vetoed falls through to the
-  previous behaviour rather than to no match at all.
+  stays eligible unless something marks it as conversation — the user's input row (`❯`/`>`),
+  sentence punctuation, a run-on past the reset clause ("…resets 9am tomorrow according to
+  the header"), or a message bullet introducing something other than a limit or reset
+  clause. The last two are what catch prose whose *wrapped* continuation happens to begin
+  with the clause. Eligibility is decided per line and never by inspecting a neighbour, so
+  the bottom-up scan — and therefore freshness — is unchanged, and anything vetoed falls
+  through to the previous behaviour rather than to no match at all.
+
+  Every signal except the prompt glyph is subordinate to whether the line **names a limit**:
+  a line that does is treated as a render however it is punctuated, glyphed or trailed. That
+  ordering is the correction to this change's own first three revisions, each of which
+  claimed the veto "only demotes what it can positively identify" and then demoted real
+  renders three ways — period-terminated banners ("Rate limit exceeded. Please try again in
+  5 hours."), banners with an inline hint ("· resets 5:20pm · /upgrade to increase your
+  limits"), and the API-error render whose vocabulary is underscored (`rate_limit_error`).
+  The two errors do not cost the same: returning prose parses a *short* wait, which is a
+  fallback and so gets revisited (#70), while returning a stale banner parses a *long* one
+  that latches non-correctable. Doubt therefore resolves toward the fresher line.
+
+  **Known boundary:** prose that names a limit *itself* — "⏺ You've hit your usage limit, so
+  try again in 2 minutes." — is per-line indistinguishable from a render and still outranks
+  the banner. This is unchanged from previous behaviour rather than introduced here, and it
+  is the recoverable direction of the two. Also unchanged: an unglyphed, unpunctuated
+  continuation that begins with the clause and ends within two words (`│  try again in 20
+  minutes  │`), which cannot be separated from a wrapped banner line (`  resets 8:40pm
+  (Europe/London)`) without dropping the only line carrying the time.
 - **The org/monthly spend-limit banner is now detected (#71).** Team/org accounts (and
   individual accounts whose extra-usage budget is exhausted) get "You've hit your org's
   monthly spend limit · run /usage-credits …" — undetected for two independent reasons:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
+- **A live banner is no longer outranked by reset-shaped prose below it (#73).** The
+  reset-message scan returned the lowest reset-shaped line in the pane, and `try again in
+  …` matches ordinary English — so a model sentence ("The API said to try again in 2
+  minutes …") or the user's own question, rendered *below* the banner it discusses, stole
+  the parse and turned a 5-hour wait into ~3 minutes. The monitor then woke into the
+  still-live limit and burned `maxRetries` before the real reset. The discriminator is
+  position within the line, not vocabulary: Claude Code *renders* a banner's limit and
+  reset clauses as the leading content of a line (behind at most the `⎿`/`└`/`⚠`/`·` echo
+  glyphs), while prose carries the same words mid-sentence — the discipline the
+  spend-limit banner already uses, generalised. Lines that merely mention a limit are now
+  skipped over, so the banner above them wins. Eligibility is decided per line and never
+  by inspecting a neighbour, which keeps the bottom-up scan — and therefore freshness —
+  exactly as it was: a stale banner still loses to any newer one below it. Anything the
+  scan does not recognise as a render falls through to the previous behaviour rather than
+  to no match at all.
 - **The org/monthly spend-limit banner is now detected (#71).** Team/org accounts (and
   individual accounts whose extra-usage budget is exhausted) get "You've hit your org's
   monthly spend limit · run /usage-credits …" — undetected for two independent reasons:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,10 +15,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   the parse and turned a 5-hour wait into ~3 minutes. The monitor then woke into the
   still-live limit and burned `maxRetries` before the real reset. The discriminator is the
   shape of the line, not its vocabulary, and it is stated as a *veto*: a reset-shaped line
-  stays eligible unless it carries a message/prompt glyph (`⏺`/`●`/`❯`) or a sentence
-  running on past the reset clause ("…try again in 2 minutes, so I'll wait"), which is what
-  catches prose whose wrapped continuation happens to begin with the clause. Lines that
-  merely mention a limit are vetoed, so the banner above them wins. Eligibility is decided
+  stays eligible unless it is the user's input row (`❯`/`>`), ends in sentence punctuation,
+  runs on past the reset clause ("…resets 9am tomorrow according to the header"), or sits on
+  a message bullet introducing something other than a limit or reset clause. The last two
+  are what catch prose whose *wrapped* continuation happens to begin with the clause. Lines
+  that merely mention a limit are vetoed, so the banner above them wins. Eligibility is decided
   per line and never by inspecting a neighbour, which keeps the bottom-up scan — and
   therefore freshness — as it was: a stale banner still loses to any newer render below it,
   including renders this file does not recognise. Anything vetoed falls through to the
--- a/src/patterns.js
+++ b/src/patterns.js
@@ -153,25 +153,54 @@ const RESET_PATTERNS = [
   /try again in \d+\s*(?:hours?|minutes?|h|m)/i,               // "try again in 5 hours"
 ];
 
-// --- Render shapes vs prose (#73) ---
+// --- Renders vs prose (#73) ---
 // A pane line may MENTION a limit or a reset time without BEING one. `/try again in/i` is
 // plain English, so "The API said to try again in 2 minutes before the window rolls" — or
 // the user's own "it told me to try again in 2 minutes, is that right?" — is reset-shaped,
 // and renders BELOW the banner it discusses, where a bottom-up scan finds it first.
 //
-// The discriminator is position within the line, not vocabulary: Claude Code RENDERS these
-// as the leading content of a line (behind at most the ⎿/└/⚠/· echo glyphs it prefixes
-// banners with), while prose carries the same words mid-sentence. This is the discipline
-// SPEND_LIMIT already uses for #71, generalised to the two halves of a banner.
+// The discriminator is the SHAPE of the line, not its vocabulary, and it is stated as a
+// VETO: a reset-shaped line is eligible unless something marks it as conversation. Stated
+// the other way round — an allowlist of known banner shapes — every render this file does
+// not recognise is demoted, including a LIVE one sitting below a stale banner, which is
+// worse than the pre-#73 behavior it replaces ("Claude usage limit reached. Resets at 2pm"
+// below "… resets 11:30am" would hand the monitor the stale time, and #70's latch would
+// then mark it non-correctable). A veto only ever demotes lines it can positively identify,
+// so an unrecognised render keeps the freshness ordering it had before.
 //
-// Deliberately NOT anchored to the message-bullet glyphs (⏺/●) or the prompt glyph (❯):
-// those introduce model output and user input respectively — exactly the lines #73 is
-// about — whereas a banner render uses the echo glyphs or no prefix at all.
-const RENDER_GLYPHS = '(?:[⎿└⚠·•]\\s*)*';
-const LINE_LEADS_WITH_LIMIT = new RegExp(
-  `^\\s*${RENDER_GLYPHS}(?:(?:you'?ve\\s+)?(?:hit|exceeded|reached)\\s+(?:your|the)\\s|\\d+-hour limit|(?:usage|rate)\\s+limit\\b)`, 'i');
-const LINE_LEADS_WITH_RESET = new RegExp(
-  `^\\s*${RENDER_GLYPHS}(?:please\\s+)?(?:resets?\\b|try again in\\b)`, 'i');
+// Two per-line signals, no neighbour lookups:
+//
+//   1. THE MESSAGE GLYPHS. ⏺/● introduce model output and ❯/> the user's own input —
+//      exactly the lines #73 is about. A banner renders behind the echo glyphs (⎿/└/⚠/·)
+//      or with no prefix at all, so it is never marked.
+//   2. WHAT FOLLOWS THE RESET CLAUSE. A render stops at the time, give or take a timezone
+//      qualifier ("· resets 3pm (UTC)", "resets 3am NY", "Please try again in 5 hours").
+//      A sentence keeps going: "…try again in 2 minutes, so I'll wait for that",
+//      "…resets 9am tomorrow according to the header". This is the half that catches prose
+//      whose WRAPPED continuation happens to begin with the clause — tmux capture-pane is
+//      called without -J (src/tmux.js), so a wrap arrives as its own line and a rule keyed
+//      on what LEADS the line sees "  try again in 2 minutes, so I'll wait" as a render.
+const MESSAGE_GLYPH = /^\s*[⏺●❯>]\s/;
+const RESET_TAIL_WORDS = 2;                          // "(Europe/London)" → 0, "NY" → 1
+
+// Text after the reset clause, or null when the line carries no reset time at all.
+function resetClauseTail(line) {
+  let end = -1;
+  for (const p of RESET_PATTERNS) {
+    const m = p.exec(line);
+    if (m) end = Math.max(end, m.index + m[0].length);
+  }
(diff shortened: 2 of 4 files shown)


nyxaria:
Pushed `d0a0e16`: the bullet exemption was demoting real renders, plus one case left open as `it.todo`.

## The bullet exemption was an allowlist

`BULLET_LEADS_WITH_CLAUSE` spelled the limit phrasings out a second time, so it recognised only what it enumerated. The comment above it argued this was safe — "an exemption inside the veto … can only ever RESCUE a line, so it carries none of the allowlist's freshness risk: a render it fails to recognise is judged by signals 2 and 3 as before." That premise is what hid the bug. The other two signals don't fire on a bulleted render, so a render the exemption misses isn't judged as before — it's **demoted, and the stale banner above it wins**.

Four shapes it forgot, each a phrase `LIMIT_PATTERNS` already carries, and each one `master` returns correctly:

```
● Claude usage limit reached · resets 2pm      /limit reached/, /usage limit/
● Session limit reached · resets 9pm           /limit reached/
⏺ Your 5-hour limit resets 3pm                 /\d+-hour limit/ — not bullet-leading
● You're out of extra usage · resets 3pm       /out of.*usage/
```

The bullet is the only discriminator in every case — strip it and the same line wins, which the new tests assert alongside the demotion. Downstream this is the bad failure, not a cosmetic one: the stale time *parses*, so `_waitIsFallback` is false (`monitor.js:111`) and only fallbacks are correctable (`:139`) — the wrong wait latches, and the monitor wakes early into a still-live limit.

The fix asks `LIMIT_PATTERNS` instead of restating it, so a render the file learns to recognise anywhere is recognised here too. `try again in` is excluded — it's a reset clause in limit clothing (`⏺ I will try again in 3 minutes` is prose) and must not vouch for a line alone.

## Open, recorded as `it.todo`: a full stop demotes a render

`SENTENCE_END` vetoes any reset-shaped line ending in `.`/`?`/`!`, but renders do end in a full stop — this file's own #71 fixture is `"You've hit your monthly spend limit."`. So both of these lose to a stale banner above them:

```
You've hit your session limit · resets 5:20pm.
● You've hit your session limit · resets 5:20pm.
```

**I did not fix this, deliberately — there's no free answer.** `SENTENCE_END` is load-bearing: dropping it un-vetoes `⏺ You've hit your usage limit, so try again in 2 minutes.` and the wrapped-prose case, both already pinned in this suite, and both of which are renders by every other available signal. Separating them needs a discriminator that doesn't exist yet, not a tweak — so it's recorded as a known demotion rather than papered over. The test is executable and one edit from live; flip `it.todo(` back to `it(` when a signal turns up.

## Where the branch stands

Measured against `master` on the same shapes — prose it correctly vetoes vs. real renders it demotes:

|  | before `d0a0e16` | after |
|---|---|---|
| prose cases fixed vs `master` | 3 | 3 |
| real renders demoted vs `master` | 7 | 2 |

Suite: **495 pass, 0 fail, 1 todo** (exit 0).

Two things I left alone, both deliberate:

- **`> resets 4:30pm (UTC)` is still demoted.** `PROMPT_GLYPH`'s no-space form is a deliberate call with an existing test pinning it, and I have no real render starting with `>` to justify overturning it.
- **Unpunctuated prose with no glyph** (`the limit resets 3pm next week`, `so it resets 5pm then`) is still not vetoed — but `master` gets these wrong too, so it's a scope limit of #73's fix rather than a regression here.

cheapestinference (changes requested):
Thanks — and the adversarial rejection of my tiered sketch is accepted in full. Your three counterexamples are correct, the retraction note is appreciated, and per-line eligibility stated as a veto is the right shape. The corpus/mutation methodology is exactly what I want to see on this file.

But the safety invariant this PR rests on — *"a veto only ever demotes what it can positively identify, so an unrecognised render keeps the freshness ordering it had before"* — does not hold as shipped. I reproduced three master-vs-PR behavioral diffs on `d0a0e16`, each the same failure you cite for rejecting both the tiered sketch and your own first cut: a real render demoted, a stale banner above it wins, and — because the stale line parses — `_waitIsFallback = false`, so #70's `correctUsageWait` refuses to revisit it.

**1. `SENTENCE_END` demotes period-terminated renders.**

```
You've hit your limit · resets 11:30am (UTC)        ← stale
some ordinary output line
Rate limit exceeded. Please try again in 5 hours.   ← live
master → "Rate limit exceeded. Please try again in 5 hours."
PR     → "You've hit your limit · resets 11:30am (UTC)"     (rolls ~21h, latched)
```

Same result for the historic CLI wording `Claude usage limit reached. Your limit will reset at 6pm (Asia/Kolkata).` The class is acknowledged in the suite only as a non-failing `it.todo` — but it isn't a todo, it's a regression: master gets both of these right, and the suite already pins (since #71) that trailing punctuation is *not* a prose signal for the spend banner. The veto uses it as one for every other render.

**2. The bullet veto demotes the real bulleted API-error render.**

```
● API Error: 429 rate_limit_error, try again in 15 minutes
```

`NAMES_A_LIMIT`'s `/rate limit/i` doesn't match the underscored API vocabulary (`rate_limit_error`, `rate_limited`), and `BULLET_LEADS_WITH_CLAUSE` fails because `API Error:` leads the line — so under a stale banner, master returns the live 15-minute line and the PR returns the stale banner, latched. Note the PR's own pinning test for this wording ("a stale banner never outranks an unrecognised render below it") strips the `●` that real renders carry — `toolEchoMask`'s comment documents the `● API Error: …` form — which is what let this pass the suite.

**3. The tail budget vetoes limit-naming renders with a >2-word suffix.**

```
You've hit your session limit · resets 5:20pm · /upgrade to increase your limits
```

The `NAMES_A_LIMIT` rescue exists only inside the bullet veto, not the tail check, so this line is demoted under a stale banner (master returns it). This one also composes badly with #75: that branch turns inline-hint banners — `…· resets 5:20pm (UTC) · run /usage-credits to finish` — from chrome into content, and every one of them has a >2-word tail. The day both PRs land, the render #75 makes parseable is demoted by this veto whenever anything stale sits above it.

**Residual #73 surface (parity with master, so not blockers — but the PR presents the hole as closed, and the Known boundary section covers only unglyphed continuations):**

- The unpunctuated variant of the headline case: `⏺ You've hit your usage limit, try again in 2 minutes` still steals the parse (the `selfVouching` fixture passes only because it carries a trailing full stop; a wrap landing after "minutes" strips it).
- Stacked closers: `It failed (the API said “try again in 2 minutes.”)` ends `.”)`, which `SENTENCE_END`'s single optional closer can't match. `?` → `*` on the closer class fixes it without touching any render (renders have no `[.?!]` before a closer).
- `∙` is missing from `BULLET_GLYPH` though the file's own `TOOL_ECHO_HEADER` knows `[●⏺∙]`; `∙ It said to try again in 2 minutes` is eligible while the identical `⏺` line is vetoed.

**Smaller:**

- `NAMES_A_LIMIT` selects its subset by grepping regex *source* text (`!/try again in/.test(p.source)`). Reword the retry pattern once (`/try again (?:in|at)/`) and the filter silently stops excluding it. Declare named subsets (`LIMIT_NAME_PATTERNS` + `RETRY_HINT_PATTERNS`, spread into `LIMIT_PATTERNS`) — the one-source-of-truth argument your own adjacent comment makes.
- One shared message-glyph class for `BULLET_GLYPH` / `BULLET_LEADS_WITH_CLAUSE` / `TOOL_ECHO_HEADER` (and `PROMPT_GLYPH` vs the input-row entries in `CHROME_LINE`).

**What I'd like before merge:** the three regression panes above pinned as failing tests first, then whatever design makes them pass. I'm deliberately not prescribing the fix: extending the `NAMES_A_LIMIT` exemption uniformly across all three signals collapses 1 and 3 but reopens the unpunctuated-prose case above, so the tension between "names a limit ⇒ never veto" and "prose about limits is the input" is real and the design choice is yours. If the honest answer is that some of these become documented boundaries rather than fixes, that's acceptable — but then the invariant claim in the comments/CHANGELOG needs to say so, because right now it's stated three times and false three ways.

Everything above was exec-verified independently on this branch and master (`node`, side-by-side imports); suite confirmed green at 494/0, which is exactly why the pins matter.

nyxaria:
Pushed `00fb97b` (the three panes as failing tests) and `ab302d6` (the design that makes them pass).

All three reproduce. I re-derived them side-by-side against `747db77` before changing anything, and the diagnosis is worse than three bugs: the invariant itself was never load-bearing. **Being phrased as a veto does not establish "only demotes what it can positively identify."** A signal that fires on a real render demotes it whichever direction the rule is written in — an allowlist demotes what it fails to recognise, and a veto demotes what it wrongly recognises. I had been treating the phrasing as the guarantee, which is why it survived three revisions and why I wrote it three times.

## The rule now

Every signal except the prompt glyph is subordinate to whether the line **names a limit**:

```js
if (PROMPT_GLYPH.test(line)) return false;                  // absolute
if (NAMES_A_LIMIT.some((p) => p.test(line))) return true;   // rescue, ahead of the rest
if (BULLET_GLYPH.test(line) && !BULLET_LEADS_WITH_CLAUSE.test(line)) return false;
if (SENTENCE_END.test(line)) return false;
…tail budget…
```

The prose signals are evidence of *prose*, and prose is all they may veto. The prompt glyph outranks even the limit name, because the user typing about their own limit is the second half of #73's report.

**The ordering is chosen on cost, not likelihood** — which I think is the part I previously had no principle for. Returning prose commits a *short* wait; it's a fallback, so #70's `correctUsageWait` revisits it. Returning a stale banner commits a *long* one that parses, so `_waitIsFallback` is false and the monitor sits on ~24h it cannot shorten. The two errors are not symmetric, so doubt resolves toward the fresher line. That single principle collapses 1, 2 and 3 together rather than patching them.

You predicted the cost: it reopens the self-vouching case. `⏺ You've hit your usage limit, so try again in 2 minutes.` now wins again.

**I took that trade, and it is smaller than it looks:** its *unpunctuated* twin already lost on master and on `d0a0e16` — you noted this yourself under "residual surface". Keeping the punctuated one vetoed was an accident of punctuation, not a principle. So this is parity with master, in the recoverable direction, and it's now written down as a boundary in the code, the CHANGELOG and a test named `BOUNDARY:` rather than claimed away. #73's actual reported line (`⏺ The API said to try again in 2 minutes …`, which names no limit) stays fixed, as does the typed-input half.

## The rest

- **Stacked closers** — `?` → `*` on the closer class, as you suggested. No render carries `[.?!]` before a closer.
- **`∙`** — now one shared `MESSAGE_GLYPH` class with `TOOL_ECHO_HEADER`. This one bit back: widening the glyph made `BULLET_LEADS_WITH_CLAUSE` load-bearing for a `∙`-bulleted render continuation, which nothing pinned. Mutation testing caught it — see below.
- **Source-text grepping** — gone. `LIMIT_NAME_PATTERNS` + `RETRY_HINT_PATTERNS`, spread into `LIMIT_PATTERNS`.
- **`rate_limit_error`** — added to the *rescue*, not to `LIMIT_PATTERNS`. I tried it your way first (`/rate[ _-]limit/i` in the shared list) and measured it: it flips whole panes to "limited" on source lines, JSON error bodies and grep output (`const RATE_LIMIT_ERROR = 429;`). The rescue can only ever make a line eligible, so widening it can't demote anything; widening `LIMIT_PATTERNS` widens `isRateLimited`. 600-call probe confirms 0 detector diffs.

## Verification

Everything re-run against `747db77`:

| | |
|---|---|
| 792-pane corpus, scored on carrying the live banner's reset time | **792/792** (baseline 66/792, **0 regressions**) |
| detector parity — `isRateLimited`, `resumedAfterLimit`, `isRateLimitOptionsPrompt`, `isWorking`, `menuStepsToWaitOption` | **0 diffs** / 3,960 calls |
| underscored-vocabulary probe | **0 diffs** / 600 calls |
| mutation | **20/21 killed** |
| suite | **508 pass, 0 todo** |

Monitor-level, since these latch and that's the whole harm — each shape below now derives ~2h from the live render instead of ~24h from the stale banner above it, with `_waitIsFallback === false` asserted:

| pane | before | after |
|---|---|---|
| stale banner + period-terminated render | ~24h, latched | **~2h** |
| stale banner + bulleted period-terminated render | ~24h, latched | **~2h** |
| stale banner + render with inline `/upgrade` hint | ~24h, latched | **~2h** |
| stale banner + `● API Error: 429 rate_limit_error, …` | ~24h, latched | **~15min** |

**Two corrections to my own methodology, both of which changed the answer:**

1. My mutation harness scored a mutant "killed" on any suite failure — including `reconcile.test.js`'s flaky cross-process lock test. That flake was hiding **two live survivors** (`bullet clause-leading exemption removed`, `separator tokens count as words`). Scoring now runs the deterministic files only. Both survivors are real gaps in *this* round's tests and are now pinned — the first is the `∙` interaction above, the second wanted a bare-clause line, because the fixture I had (`5-hour limit reached · resets 3pm - see above`) now short-circuits on the rescue and was asserting the rescue rather than the separator handling. Several other tail-budget tests had gone vacuous the same way; those now assert both forms.
2. The one surviving mutant is a `lastIndex` advance I added to `resetClauseTail` as a guard against a future zero-length `RESET_PATTERN` spinning the loop forever. It is undetectable by construction, so I've left it in the harness labelled as an expected survivor rather than dropping it to claim 21/21. (Mutation testing found the hazard by *hanging* on the non-global mutant — `g` is load-bearing for termination there, not just for finding the last clause, and the guard does not cover that case. Comment says so.)

For the record, `reconcile.test.js`'s cross-process test fails on `747db77` too under full-suite load and passes 4/4 in isolation — pre-existing, untouched by this branch.

## On #75

The inline-hint shape is fixed by the same rescue, since it keys on the limit name rather than the tail — so the collision you flagged doesn't arise. I couldn't assert the `/usage-credits` wording directly here because `isChromeLine` still classifies it as furniture on this branch; the test uses an equivalent `/upgrade` tail and names #75 in the comment. Worth re-checking when that lands.

cheapestinference:
Heads-up: #75 just landed on master (`8c3d6b4`), so this branch's rebase will pick up the newly-content inline-hint banners — the composition test we agreed on (`…· run /usage-credits to finish` vs the tail-word budget) belongs here now, per your note on #75. The three regression panes from the review stand as before.

nyxaria:
*(Edited. This comment originally led with a merge-order recommendation — #75 had already landed when I posted it, and I'd missed your note above. Replaced with the current state rather than left to contradict you.)*

#75 is in master (`8c3d6b4`, released as v0.7.2), so the ordering question is settled, and the composition test belongs here exactly as you say. It's now on this branch:

- `8490417` — merge `upstream/master`. Clean, no conflicts, CHANGELOG included.
- `04d2aae` — the pin: `findRateLimitMessage` on `[stale "…resets 11:30am (UTC)", "● wrote some code", "You've hit your session limit · resets 5:20pm (UTC) · run /usage-credits to finish"]` must return the live line. It was written on #75's branch before the merge and cherry-picked once master carried the chrome anchoring; the `/upgrade` test stays where it is, and this one sits beside it with the real wording.

It has teeth rather than passing vacuously — verified against both heads of this branch's history: at `d0a0e16` the pane returns the stale banner and the test fails; at `ab302d6` it returns the live line and passes.

The collision itself does not materialise, which is the substance of the earlier note and still holds: `NAMES_A_LIMIT` runs ahead of the tail-word budget, so the inline-hint banner's >2-word tail never gets a vote.

Suite on `04d2aae`: 528 tests, 527 pass, 0 todo, 1 fail — the `reconcile.test.js` cross-process lock flake, which passes in isolation and fails on base too.

cheapestinference (approved):
Re-verified `04d2aae` with the same side-by-side harness as the review — every finding resolved or explicitly pinned:

```
period-terminated render        stale → live ✓     stacked closers .”)      fixed BEYOND master ✓
historic CLI wording            stale → live ✓     ∙ bullet variant         fixed beyond master ✓
● API Error 429 rate_limit_*    stale → live ✓     unglyphed continuation   fixed beyond master ✓
>2-word tail (/upgrade)         stale → live ✓     limit-naming prose       parity, pinned BOUNDARY test ✓
inline-hint banner (composition pin)   live ✓      #73 headline cases       banner wins ✓
```

The structural asks all landed too: `NAMES_A_LIMIT` is now composed from named subsets instead of grepping pattern source, the API vocabulary covers the underscored wordings, the message-glyph class is shared, and the exemption runs ahead of all three veto signals — which is what makes the invariant claim true this time. Red-first commit (`00fb97b`) pinning the three regressions before the fix is exactly what I asked for. Suite 527/0 here.

The one surviving steal — unpunctuated prose that itself names a limit — is master-parity, inherent to the exemption tension we both called out, and now visible in the suite as a BOUNDARY test rather than implied by omission. That's the honest version of the invariant. Approving and merging. This plus the reconcile lock fix will go out as v0.7.3.

PR #75, opened 2026-08-14
fix(patterns): four spend-limit render defects (#71 follow-up)

nyxaria:
Four defects in the #71 spend-limit path, found while reviewing #74. All four reproduce on `master` (v0.7.1) and are independent of #74 — the two branches touch different parts of `patterns.js`, so they can land in either order (the CHANGELOG will conflict trivially).

Three of them share a root cause: **a limit banner can name `/usage-credits` on the same line as the limit itself**, and the chrome allowlist matched that hint anywhere.

## 1. A banner naming the hint inline is classified as furniture

`USAGE_CREDITS` is used for two jobs — the live-limit *signal* (wants the hint anywhere) and the chrome *entry* (wants the companion row). The companion is not the only place it renders:

```
You've hit your session limit · resets 5:20pm (UTC) · run /usage-credits to finish
```

That line is chrome on master, so `contentTail` strips it. The limit is still detected through the live-region backstop, but the line carrying the reset time never reaches the parse:

```
master:  isRateLimited=true   findRateLimitMessage=null   -> fallbackWaitHours, latched correctable
this PR: isRateLimited=true   findRateLimitMessage=<the banner>  -> the real reset time
```

The spend banner (which *always* carries the hint inline) needed an explicit `!SPEND_LIMIT.test(l)` exemption in `isChromeLine` to be visible at all. Anchoring the chrome entry to the hint **leading** its row covers every banner that mentions it, so that exemption is gone — and with it the forward reference to a `const` declared 100 lines further down, which survives today only because nothing calls `isChromeLine` during module evaluation.

Lines that stop being furniture become content the detectors can see, so prose naming the hint mid-sentence is pinned inert: no reset time, no banner shape, fails on its own merits.

## 2 & 3. Two spend-banner shapes that were total misses

```
⚠ You've hit your org's monthly spend limit · run /usage-credits to raise it   -> master: undetected
· You've hit your org's monthly spend limit · run /usage-credits to raise it   -> master: undetected
You’ve hit your org’s monthly spend limit · run /usage-credits to raise it     -> master: undetected
```

The marker class admitted only `⎿`/`└`, though `⚠` and `·` prefix limit renders everywhere else in this file. And `you'?ve` required an ASCII apostrophe while the qualifier class beside it already admitted `’` for `org’s` — that class was widened for exactly this reason when #71 landed; the word in front of it was not.

Each is a *total* miss rather than a weaker match, because the banner also carries the hint inline: pattern rejects it → defect 1 hides it → nothing downstream sees it.

## 4. A wrapped quotation of the banner false-fires a wait

`^\s*` is not an anchor. Model output wraps with a hanging indent, so a continuation line begins with whitespace and then whatever the sentence was up to:

```
⏺ The team banner reads:
  You've hit your org's monthly spend limit · run /usage-credits to raise it
```

On master that enters a 5-hour wait and then types up to `maxRetries` messages into an idle session. The spend render has no reset time to sanity-check against, so its shape is the whole defense and has to hold.

A render starts at column 0 or behind an echo marker; indented with no marker is a wrap. A quotation that lands at column 0 still fires — the remaining anchors are the live-region gate and the bounded, correctable fallback. Trailing punctuation is deliberately *not* a signal: `You've hit your monthly spend limit.` is a real render with a full stop, pinned in the suite since #71.

## Verification

Differential over 4,913 three-line panes built from the spend/companion/prose/banner shapes, master vs this branch — every diff is one of the four fixes above:

```
252x  findRateLimitMessage null -> the banner   (inline-hint session banner, defect 1)
181x  isRateLimited false -> true               (⚠-prefixed render)
181x  isRateLimited false -> true               (·-prefixed render)
181x  isRateLimited false -> true               (typographic apostrophes)
133x  isRateLimited true -> false               (wrapped quotation, defect 4)
```

No new false positives: prose naming the hint mid-sentence, prose explaining spend limits, and a stale spend banner with real work below it all stay inert. A separate 4,913-pane corpus over the *unchanged* shapes (`isRateLimited`, `resumedAfterLimit`, `isRateLimitOptionsPrompt`, `isWorking`, `menuStepsToWaitOption`) shows **0 behavioural diffs** across 24,565 calls.

Suite: **463 pass, 0 fail**. (`reconcile.test.js`'s cross-process hold test is flaky on master too — ~50% locally, unrelated.)

## Commits

1. `fb39ad0` — the `/usage-credits` hint is furniture only when it leads its row.
2. `667fd68` — the spend banner's marker and apostrophe shapes were misses.
3. `dfd6d3e` — a wrapped quotation of the spend banner is not a render.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,20 @@ All notable changes to this project are documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+- **A banner that names `/usage-credits` inline is no longer treated as UI furniture.** The
+  chrome allowlist matched the hint anywhere on a line, but the companion row is not the
+  only place it renders — a session banner can carry it inline ("You've hit your session
+  limit · resets 5:20pm · run /usage-credits to finish"), and the spend-limit banner always
+  does. Those lines were stripped as chrome, so the tail dropped the only line naming the
+  limit: the spend banner needed a special case to be seen at all, and a session banner's
+  reset time was invisible to the parse, sending the monitor to the `fallbackWaitHours`
+  default instead of the real reset. The entry is now anchored to the hint *leading* its
+  row, which covers every banner that mentions it — and removes both the special case and
+  its forward reference to a constant declared 100 lines below.
+
 ## [0.7.1] - 2026-08-14
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,19 +20,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   its forward reference to a constant declared 100 lines below.
 - **Spend-limit render shapes that were total misses (#71 follow-up).** The banner pattern
   admitted only the `⎿`/`└` echo markers, so the `⚠`- and `·`-prefixed renders — markers
-  this file already expects on limit banners — were never matched; and it required an ASCII
-  apostrophe in "you've" while the qualifier beside it already admitted `’`, so a render in
-  typographic quotes was missed too. Both compounded with the chrome misclassification
-  above: invisible rather than merely unanchored.
+  this file already expects on limit banners — were never matched; the `⚠` in emoji
+  presentation (`⚠️`, U+26A0 + U+FE0F) failed on the variation selector; the **boxed** form
+  (`│ ⚠ You've hit … │`, the render this suite already pins for session banners) was not
+  admitted at all, and unlike a session banner the spend render has no reset line to fall
+  back on; and the pattern required an ASCII apostrophe in "you've" while the qualifier
+  beside it already admitted `’`, so a render in typographic quotes was missed too. All of
+  them compounded with the chrome misclassification above: invisible rather than merely
+  unanchored. The apostrophe is now *required* in the other direction — "youve hit your
+  monthly spend limit" no longer walks the one detection path that needs no reset time.
 - **A wrapped or bulleted quotation of the spend banner no longer false-fires a wait.**
   `^\s*` is not an anchor when model output wraps with a hanging indent: a continuation line
   beginning "You've hit your org's monthly spend limit …" entered a 5-hour wait and typed
-  retries into an idle session. A render starts at column 0, behind a flush-left `⚠`/`·`
-  banner marker, or behind a `⎿`/`└` tool-echo marker (the only ones that legitimately render
-  indented, as children of the notice above them). An indented `⚠`/`·`/`•` is a prose bullet
-  — the shape a model quoting the banner actually produces — and no longer counts as a
-  render. (Trailing punctuation deliberately stays out of it — "You've hit your monthly spend
-  limit." is a real render with a full stop.)
+  retries into an idle session. A render starts at column 0, behind a box border, behind a
+  flush-left `⚠`/`·` banner marker, or behind a `⎿`/`└` tool-echo marker (the only ones that
+  legitimately render indented, as children of the notice above them). An indented `⚠`/`·`/`•`
+  is a prose bullet — the shape a model quoting the banner actually produces — and no longer
+  counts as a render. (Trailing punctuation deliberately stays out of it — "You've hit your
+  monthly spend limit." is a real render with a full stop.) The column-0 rule is a rendering
+  assumption, so it has one escape hatch: when the **standalone** `/usage-credits` companion
+  row renders below the banner — evidence a quotation essentially never carries, since it
+  reproduces the banner line and not the separate row beneath it — the indented form is
+  accepted, keeping a real render printed one space too far right, or wrapped.
 
 ## [0.7.1] - 2026-08-14
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,12 +24,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   apostrophe in "you've" while the qualifier beside it already admitted `’`, so a render in
   typographic quotes was missed too. Both compounded with the chrome misclassification
   above: invisible rather than merely unanchored.
-- **A wrapped quotation of the spend banner no longer false-fires a wait.** `^\s*` is not an
-  anchor when model output wraps with a hanging indent: a continuation line beginning
-  "You've hit your org's monthly spend limit …" entered a 5-hour wait and typed retries into
-  an idle session. A render starts at column 0 or behind an echo marker; indented with no
-  marker is a wrap. (Trailing punctuation deliberately stays out of it — "You've hit your
-  monthly spend limit." is a real render with a full stop.)
+- **A wrapped or bulleted quotation of the spend banner no longer false-fires a wait.**
+  `^\s*` is not an anchor when model output wraps with a hanging indent: a continuation line
+  beginning "You've hit your org's monthly spend limit …" entered a 5-hour wait and typed
+  retries into an idle session. A render starts at column 0, behind a flush-left `⚠`/`·`
+  banner marker, or behind a `⎿`/`└` tool-echo marker (the only ones that legitimately render
+  indented, as children of the notice above them). An indented `⚠`/`·`/`•` is a prose bullet
+  — the shape a model quoting the banner actually produces — and no longer counts as a
+  render. (Trailing punctuation deliberately stays out of it — "You've hit your monthly spend
+  limit." is a real render with a full stop.)
 
 ## [0.7.1] - 2026-08-14
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   apostrophe in "you've" while the qualifier beside it already admitted `’`, so a render in
   typographic quotes was missed too. Both compounded with the chrome misclassification
   above: invisible rather than merely unanchored.
+- **A wrapped quotation of the spend banner no longer false-fires a wait.** `^\s*` is not an
+  anchor when model output wraps with a hanging indent: a continuation line beginning
+  "You've hit your org's monthly spend limit …" entered a 5-hour wait and typed retries into
+  an idle session. A render starts at column 0 or behind an echo marker; indented with no
+  marker is a wrap. (Trailing punctuation deliberately stays out of it — "You've hit your
+  monthly spend limit." is a real render with a full stop.)
 
 ## [0.7.1] - 2026-08-14
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   default instead of the real reset. The entry is now anchored to the hint *leading* its
   row, which covers every banner that mentions it — and removes both the special case and
   its forward reference to a constant declared 100 lines below.
+- **Spend-limit render shapes that were total misses (#71 follow-up).** The banner pattern
+  admitted only the `⎿`/`└` echo markers, so the `⚠`- and `·`-prefixed renders — markers
+  this file already expects on limit banners — were never matched; and it required an ASCII
+  apostrophe in "you've" while the qualifier beside it already admitted `’`, so a render in
+  typographic quotes was missed too. Both compounded with the chrome misclassification
+  above: invisible rather than merely unanchored.
 
 ## [0.7.1] - 2026-08-14
 
--- a/src/patterns.js
+++ b/src/patterns.js
@@ -29,8 +29,16 @@ const USAGE_CREDITS = /\/usage-credits\b/i;
 // classified those banners as furniture: contentTail stripped them, so the only line
 // naming the limit — and, for a session banner, the only line carrying the reset time —
 // was invisible to every detector. The companion always renders on its own row (indented,
-// or behind the ⎿/└/· echo marker).
-const USAGE_CREDITS_HINT = /^\s*(?:[⎿└·]\s*)*\/usage-credits\b/i;
+// or behind ONE ⎿/└/· echo marker — no render stacks them), so this is the same literal as
+// the signal above with a leading anchor, composed from it rather than re-spelled.
+//
+// Leading indentation is admitted here and REFUSED for the same `·` in SPEND_LIMIT below.
+// The two constants answer different questions about opposite failure costs: this one asks
+// "is this row furniture?" about a row the TUI renders INDENTED in every observed pane, and
+// over-matching costs at most a stripped line that carries no limit signal of its own.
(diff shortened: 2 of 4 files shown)


nyxaria:
Reviewed. The four defects hold up and the reasoning behind each is sound — I reproduced all four on `master`. One follow-up defect in commit `667fd68`, fixed in `ee6f16c` and pushed to this branch.

## The marker class reopens the hole `dfd6d3e` closes

`667fd68` widened the prefix to `(?:\s*[⎿└⚠·•])+`, which lets **any** of those glyphs carry leading whitespace. `dfd6d3e` then closes the wrap hole for a *bare* hanging indent — but not for the shape a model quoting a banner actually produces, which is a bullet:

```
⏺ Two things could be happening:
  · You've hit your org's monthly spend limit · run /usage-credits to raise it
```

On this branch before `ee6f16c` that returns `isRateLimited=true` — a 5h wait and `maxRetries` messages typed into an idle session, from prose. Same for `  • …`, `  ⚠ …`, and `• …` at column 0. It's the exact failure `dfd6d3e`'s own test pins, one bullet character away.

## Which marker may be indented is part of the shape

The two marker classes aren't alike, so they shouldn't share a rule:

- **`⎿`/`└` are tool-echo markers.** They render as indented children of the notice above them (`● Agent "…" finished` / `  ⎿ You've hit …`) — the live incident this suite already pins — so they must keep leading whitespace. Neither is a character prose reaches for, so it costs nothing.
- **`⚠`/`·` are banner markers, and the TUI renders them flush left.** Every render in `patterns.js` and its suite sits at column 0, including the sketch at the top of the detection block. Indented, they're prose bullets.

```js
/^(?:\s*[⎿└]\s*|[⚠·]\s*)?you['’]?ve\s+…/i
```

`•` is dropped outright: it appears nowhere else in the codebase, no reported render uses it, and it's the one glyph that is purely a prose bullet — it came in alongside `⚠`/`·` without a render to justify it. Same for the companion-hint entry, whose own comment enumerates `⎿`/`└`/`·`.

**Differential**, 2,000 panes × 5 detectors, `667fd68`+`dfd6d3e` vs `ee6f16c` — one behavioural class changes:

```
200x  isRateLimited true -> false     (all bulleted or indented quotations)
```

Every real shape still fires: column 0, `⚠ `, `· `, and `⎿`/`└` indented or not. Suite 463 pass / 0 fail (`reconcile.test.js`'s cross-process test is the known flake — passes in isolation).

## Two notes, no change requested

- **Not a regression, but still open:** print mode never sees this banner. `isRateLimited(spendPane, [], 0)` is `false` — the companion backstop is behind `if (tailLines > 0)`, and the main loop still wants a `RESET` line the spend render never has. Pre-existing from #71 and out of scope here, but `claude -p` under the launcher exits on a spend limit instead of retrying. Worth its own issue.
- **`●`/`⏺` deliberately stays out** of the marker class. It's the assistant-message glyph, so `● You've hit your org's monthly spend limit …` is far likelier a model quoting the banner than a render — and with no reset time to sanity-check against, it's the most dangerous marker that could be added.

I also merged this branch into #74's: `src/patterns.js` auto-merges and the combined suite is **502 pass / 0 fail**, so the independence claim checks out — only the CHANGELOG conflicts.

cheapestinference (changes requested):
Strong PR. The chrome half checks out clean from every angle I threw at it: I traced the anchored hint through `contentTailRange`, the backstop gate, `findRateLimitMessage`, `resumedAfterLimit` and the monitor's wait/correction paths, and found no realistic pane that loses detection — the inline-hint session banner going from `findRateLimitMessage = null` (5h fallback) to the real reset time is exec-confirmed, as are the ⚠/·/curly-apostrophe fixes and the wrap veto doing its job. Dropping the `!SPEND_LIMIT` exemption is safe (no remaining `CHROME_LINE` entry can match a banner shape), and the new anchors are linear and net *faster* than what they replace. Requesting changes on a short list:

**1. The marker-shape class this PR claims closed still leaves fixture-pinned renders open — and they're total misses, the failure mode your own description says is the expensive one.**

- **Boxed render:** `│ ⚠ You've hit your org's monthly spend limit · run /usage-credits to raise it │`. The suite pins the boxed form for session banners (`'│ ⚠ You've hit your limit │'`, patterns.test.js:526) — those survive via the unanchored LIMIT+RESET pair, but the spend banner has no reset line, so `SPEND_LIMIT` is its only path and neither alternative admits `│` (nor a space-prefixed `⚠`). `isRateLimited` = false, monitor idles forever.
- **Emoji presentation:** `⚠️` is U+26A0 + U+FE0F; the variation selector is neither `\s` nor `y`, so the whole pattern fails. `[⚠·]\uFE0F?\s*` fixes it.
- **Branch asymmetry:** ` ⚠ You've …` with a single leading space fails while `  ⎿  You've …` passes (`\s*[⎿└]\s*` vs bare `[⚠·]\s*`). One rendering margin away from a miss.

Either widen the pattern or pin these as explicit known-boundary tests — right now the comment block says "every render in this file and its suite sits at column 0" while the boxed fixture 300 lines down contradicts it.

**2. The indented-markerless regression deserves one refinement.** Exec-confirmed: `"  You've hit your monthly spend limit."` flips from detected (master) to missed — *even when the genuine `/usage-credits` companion row renders directly below it*. The trade-off itself is sound and disclosed, but the companion row in the live region is a liveness signal a model-quoted wrap essentially never brings with it. Consider accepting the indented markerless form iff the standalone companion row is present below; that keeps the wrap veto for the quoted case and removes most of the residual risk of the "flush-left always" assumption. Suggestion, not a blocker — if you'd rather keep the rule pure, pin the trade-off with a test comment saying what was given up.

**3. Reuse/consistency:**

- `USAGE_CREDITS_HINT` re-inlines the `/usage-credits\b` literal, and the comment above `USAGE_CREDITS` still says "one source of truth for both" — now false. Compose it: `new RegExp('^\\s*(?:[⎿└·]\\s*)*' + USAGE_CREDITS.source, 'i')`.
- `(?:[⎿└·]\s*)*` — the star admits stacked markers no render produces; `?` says what you mean.
- `·` now carries opposite indentation semantics in the two constants this PR touches (`USAGE_CREDITS_HINT`: indented `·` OK; `SPEND_LIMIT`: indented `·` is a prose bullet). Both are defensible in isolation (the hint anchor is still strictly narrower than master's match-anywhere), but nothing ties them together — align them or cross-reference why they differ.
- `you['’]?ve` — the rewrite was the cheap moment to make the apostrophe required; `youve hit your monthly spend limit` at column 0 still satisfies the only no-reset-anchor path into a wait.

**4. Noting, no change requested:** a *stale* inline-hint banner now parses to a real time, so an already-resumed pane can commit a date-rolled ~22h wait, latched non-correctable, where master took the correctable 5h fallback. It's the known stale-banner hazard extended to newly visible lines (any user activity scrolls it away), and the fix isn't in scope here — flagging it so it's on record. Also: heads-up that #74's tail-word budget would demote exactly the inline-hint banner this PR makes parseable (`· run /usage-credits to finish` is a >2-word tail) — flagged on that PR too; whichever lands second needs a composition test.

Verification notes: all of the above exec-verified on the branch head vs master (side-by-side imports), suite green at 463/0 here too, and your differential-corpus claims reproduced for every case I sampled. The four fixes themselves are real and wanted — with #1 and #3 addressed this merges quickly.

nyxaria:
All of it addressed in `f7fb6f3`. Every claim reproduced on the branch head first — the boxed form, the variation selector, the one-space margin and the `youve` path all behave exactly as described:

```
false  │ ⚠ You've hit …          false  ⚠️ (U+26A0 U+FE0F)
false  ' ⚠ ' one space margin    false  indented markerless + genuine companion row
true   'youve hit your monthly spend limit'  ->  reaches the no-reset-anchor wait
```

## 1. Widened, not pinned as boundaries

All three were total misses, so pinning them as known-boundary tests would have been documenting a hole rather than closing one. The prefix is now three alternatives instead of two:

```js
`^(?:\\s*[⎿└]\\s*|\\s*│\\s*(?:[⚠·]\\uFE0F?\\s*)?|[⚠·]\\uFE0F?\\s*)?` + SPEND_LIMIT_BODY.source
```

`│` is a **box border**, not a marker, and that is why it may be indented where a bare `⚠` may not: a box is indented or not *as a whole*, so it carries its own column-0 evidence. The comment block says that now instead of the "every render sits at column 0" claim your boxed fixture contradicted — you were right that the fixture and the comment could not both be true.

## 2. Implemented, not just documented

The one-space margin and the indented-markerless regression are the same defect wearing two hats: a rendering assumption about the left edge, whose failure mode is a total miss. Rather than trade one against the other, `SPEND_LIMIT_INDENTED` drops the column-0 rule entirely and is consulted **only when the pane's companion is the standalone row** rather than a hint the banner names inline. Since `companionIdx` is already the last `/usage-credits` line, "the companion is its own row" is exactly `USAGE_CREDITS_HINT` applied to it — no new scan:

```js
const spendShapes = USAGE_CREDITS_HINT.test(all[companionIdx])
  ? [SPEND_LIMIT, SPEND_LIMIT_INDENTED] : [SPEND_LIMIT];
```

The bulleted quotation stays inert because the banner names the hint inline, so the last companion match *is* the quoted line and the hatch never opens. Prose bullet glyphs (`•`, `-`, `*`) stay out even when it does — no render prints them at any indent — and `●`/`⏺` stays out everywhere, for the reason you gave.

The residual cost is a quotation that reproduces the companion row too:

```
⏺ The team banner reads:
  You've hit your org's monthly spend limit · run /usage-credits to raise it
     /usage-credits to finish what you're working on.
```

That fires. Pinned as an explicit `KNOWN BOUNDARY:` test with the reasoning — at the shape level it is indistinguishable from the render, the live-region gate is what remains, and the wait it produces is the bounded correctable fallback.

## 3. Reuse

- `USAGE_CREDITS_HINT` composes from `USAGE_CREDITS.source` — "one source of truth for both" is true again.
- `*` → `?`. No render stacks markers.
- `SPEND_LIMIT` / `SPEND_LIMIT_INDENTED` compose from one `SPEND_LIMIT_BODY`.
- The `·` asymmetry is written down at **both** constants rather than aligned, because the two answer different questions with opposite failure costs: the hint asks "is this row furniture?" about a row the TUI renders *indented* in every observed pane, where over-matching costs a stripped line carrying no limit signal; `SPEND_LIMIT` asks "is this row a live render?" about a line printed flush left, where over-matching commits a multi-hour wait with nothing to check it against.
- Apostrophe now **required**. Straightforward tightening: every real render prints one.

## 4. Noted

Both on record, neither in this PR. The stale-banner hazard is real and pre-existing — this PR widens which lines it can reach, not the hazard itself. On #74: agreed that whichever lands second needs the composition test for the inline-hint banner vs the tail-word budget; I'll write it into the second one.

## Verification

Differential over 6,480 panes × 5 detectors (32,400 calls), previous head vs `f7fb6f3` — two behavioural classes, both intended:

```
896x  isRateLimited false -> true    (boxed │, ⚠️ emoji presentation, indented + standalone companion)
160x  isRateLimited true -> false    ("youve", apostrophe now required)
```

Broken out by prefix, the `false -> true` set is exactly `│ `, `│ ⚠ `, `⚠️`, `⚠️ ` at any indent, plus `  `, ` ⚠ `, ` · `, `  · ` **only** with the standalone companion row present — never without it. Zero diffs on `findRateLimitMessage`, `resumedAfterLimit`, `isRateLimitOptionsPrompt`, `isWorking`. Against master the only losses remain the `youve` panes and the wrap veto itself.

Suite: 470 pass / 1 fail — `reconcile.test.js`'s cross-process hold test, which I reproduced failing on the PR base `70d810a` (1 of 3 runs there, 3 of 3 in the worktree under load). Nothing in this branch touches the lock path.

cheapestinference (approved):
Verified `f7fb6f3` independently — every claim reproduces exactly as described (side-by-side master vs branch, same harness as the previous round):

```
boxed │ ⚠ …                  false → true      indented, NO companion      true → false  (intended)
⚠️ (U+26A0 U+FE0F)           false → true      wrapped/bulleted quotation  true → false  (intended)
one-space ⚠ margin           false → true      youve (no apostrophe)       true → false  (intended)
indented + companion row     true  → true      known-boundary quote+companion pinned, fires as documented
curly apostrophes            false → true      inline-hint banner parse    null → real reset time
```

Suite green here too (470/0 — the `reconcile` cross-process hold flake didn't even trigger on this machine, and it reproduces on the PR base, so it's not this branch's). The companion-gated hatch is a better design than what I asked for — reusing `companionIdx` + `USAGE_CREDITS_HINT` as the gate means no new scan and the bulleted-quote case stays closed for free. The `·` asymmetry written down at both constants with the opposing failure costs is the right call over forced alignment.

Approving and merging. The #74 composition test lands with #74 as agreed.

nyxaria:
## Merge order: this one first, then #74

The two branches are independent in code — `src/patterns.js` auto-merges, only the CHANGELOG conflicts — so this is a preference rather than a constraint. Verified on `ab302d6` + `f7fb6f3`: combined suite **526 pass / 1 fail** (the `reconcile.test.js` cross-process flake, which also fails on base `70d810a`), no other conflict.

The reason to take this one first is the composition test you asked for — item 4 here, item 3 on #74:

- **It can only be written on this branch.** On #74 alone, `isChromeLine` still classifies `…· run /usage-credits to finish` as furniture, so `findRateLimitMessage` never sees the line and the assertion is unwritable — which is why #74 pins an equivalent `/upgrade` tail and defers the real wording.
- **Merging this first puts the guard in master before the veto arrives**, so #74's merge is checked by it. In the other order the tail-word budget sits in master with nothing pinning the interaction. Harmless in itself — nothing turns inline-hint banners into content until this branch lands — but the pin would arrive alongside the change it exists to guard against rather than ahead of it.

The secondary reason: #74 is the larger design change and is still awaiting your re-review of `ab302d6`, while this branch's remaining scope is one test.

## The collision itself does not materialise

Re-checked on the merge rather than assumed, on the pane the heads-up describes:

```
[ "You've hit your limit · resets 11:30am (UTC)",              ← stale
  "● wrote some code",
  "You've hit your session limit · resets 5:20pm (UTC) · run /usage-credits to finish",   ← live
  "", "❯ " ]

#75 alone (f7fb6f3)   ->  the live 5:20pm line
#74 alone (ab302d6)   ->  the stale banner      (expected — the hint is still chrome there)
merged                ->  the live 5:20pm line
```

#74's `NAMES_A_LIMIT` rescue runs ahead of the tail-word budget, so a line naming a limit is never demoted for tail length — the >2-word tail on the inline-hint banner never gets a vote. The pin for it lands on this branch shortly.

Flagging it here rather than duplicating the same note on #74.

cheapestinference:
Follow-up on the `reconcile.test.js` flake you flagged: it was a real race, not a bad test. After winning the breaker, the stale-lock removal was an unconditional `unlink` even when the re-read found the lock already absent — a fast-path acquirer could create a fresh live lock in the read→unlink gap and the unlink deleted it → double-hold. Reproduced at ~0.2% of contended rounds on a single loaded core with an acquisition-path event log; fixed on master in `4d7071f` (identity-checked stale removal, skipped when absent — 0 overlaps in 6,205 holds under the same load after). Your ~50% local rate under load now makes sense: the test was a statistical canary doing its job. Thanks for reporting it.

PR #77, opened 2026-08-17
feat(monitor): resume a turn truncated by suspend or a dropped connection

nyxaria:
## The gap

Close your laptop lid while Claude is mid-answer and the work does not survive. Claude Code wraps the response body in a byte watchdog; when the bytes stop arriving it aborts the stream and finalizes whatever had already been printed, naming the cause on an `API Error:` line:

```
⏺ API Error: Your computer went to sleep mid-response. The response above may be
  incomplete.
```

The turn is then **over**. The prompt returns idle and nothing resumes it, so the session can sit on a half-finished answer indefinitely — the same shape of stall this tool already fixes for usage limits, overloads and safeguard flags.

It cannot be waited out. Claude Code retries by itself only while the response is still *thinking-only*; once real content has been yielded it declines to retry, which is exactly when this render appears. Seeing it at an idle prompt, the monitor now sends `continue`, bounded at `maxRetries` because a machine that just woke may not have its network back yet.

## Scope: seven renders, one remedy

The same finalizer emits the render for a dropped connection, a stalled stream and a mid-response server error, in both "mid-response" and "before a response was produced" forms. All leave the identical truncated-turn state and take the identical remedy, so all are matched — sleeping is just the cause we can name. Matching only the suspend wording would miss the same event on a narrow pane or a different provider path.

## Anchoring differs from the other families, deliberately

The overload and safeguard paths require an `API Error` line *nearby*. That works because their phrases are jargon. These are ordinary English about a common event, so a session merely **explaining** the error quotes the whole render — anchor included — mid-sentence, and a pane discussing this feature would inject into itself.

The discriminator is therefore the **shape** of the line, the same veto discipline as the render/prose split in the reset path: a real render *begins* with `API Error:`, behind at most Claude's message glyph. Prose carries it mid-line; the user's own line carries `❯`. Both are covered by tests.

## Verification

Verified end to end against a **real physical suspend**, not only fixtures. With `CLAUDE_BYTE_STREAM_IDLE_TIMEOUT_MS=30000` to shorten the watchdog's 180s first-party idle deadline, an 85s sleep mid-stream produced the render, and the monitor detected it 11 seconds after wake and resumed the turn with a single `continue`:

```
[WARN] Interrupted stream detected [matched /went to sleep mid-response/ in:
       "⏺ API Error: Your computer went to sleep mid-response. The response above"]
       — the turn was truncated and left at an idle prompt. Will resume up to 2x every 5s.
[INFO] Resume sent after interrupted stream (attempt 1/2).
```

A separate run against a static pane confirmed the bound: exactly 2 sends, then a loud give-up, and it stops typing.

**25 new tests** in `test/stream-interrupted.test.js` (pattern matching, config validation, and the full `processOneTick` state machine), written test-first. Suite is 553/553 green on this branch.

The false-positive cases are pinned explicitly: prose quoting the render mid-sentence, the phrase typed by the user, the render quoted inside a tool result, the render left far up in scrollback, and a healthy session discussing the error at an idle prompt.

## Notes

- Config lives under a `streamInterrupted` block (`enabled`, `patterns`, `maxRetries: 2`, `retryDelaySeconds: 5`, `retryMessage: "continue"`), documented in README with the other families.
- Config validation gained `validateBoundedRetry` / `validPatterns`, now shared by the safeguard and overload blocks rather than a third copy of the same shape.
- The monitor branch is kept parallel to the safeguard one rather than merged, so each family's result labels stay greppable from the log and status-file code that matches them literally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,27 @@ All notable changes to this project are documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+- **A turn truncated by a suspended machine or a dropped connection is now resumed.**
+  Claude Code wraps the response body in a byte watchdog; when the bytes stop arriving it
+  aborts the stream and finalizes whatever had already been printed, naming the cause on an
+  `API Error:` line ("Your computer went to sleep mid-response. The response above may be
+  incomplete."). The turn is then over — the prompt returns idle and nothing resumes it, so
+  a session can sit on a half-finished answer indefinitely. Claude Code retries by itself
+  only while the response is still thinking-only; once real content has been yielded it
+  declines to retry, which is precisely when this render appears. Seeing it at an idle
+  prompt, the monitor now sends `continue`, bounded at `maxRetries` because a machine that
+  just woke may not have its network back yet. All seven renders the finalizer emits are
+  matched (suspend, dropped connection, stalled stream and mid-response server error, in
+  both their "mid-response" and "before a response was produced" forms) — they leave the
+  same wreckage and take the same remedy. Detection is anchored on the SHAPE of the line
+  rather than its vocabulary: a real render *begins* with `API Error:`, behind at most
+  Claude's message glyph, so a session merely explaining the error — which quotes the whole
+  render, anchor included, mid-sentence — cannot trigger a resume. Configured under a
+  `streamInterrupted` block.
+
 ## [0.7.3] - 2026-08-16
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,8 +23,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   same wreckage and take the same remedy. Detection is anchored on the SHAPE of the line
   rather than its vocabulary: a real render *begins* with `API Error:`, behind at most
   Claude's message glyph, so a session merely explaining the error — which quotes the whole
-  render, anchor included, mid-sentence — cannot trigger a resume. Configured under a
-  `streamInterrupted` block.
+  render, anchor included, mid-sentence — cannot trigger a resume. The glyph and the
+  indentation are one rule: a glyphed head may sit anywhere, a bare head must start at
+  column 0, because an indented glyph-less head is the hanging-indent shape of a quotation
+  rather than a render. Configured under a `streamInterrupted` block.
 
 ## [0.7.3] - 2026-08-16
 
--- a/README.md
+++ b/README.md
@@ -83,6 +83,7 @@ When you disconnect (SSH drops, close terminal, laptop sleeps), **tmux keeps run
 - **Self-healing coverage** — `reconcile` re-arms monitors for any live `claude` session that lost one; an optional timer (`systemd --user` on Linux, launchd on macOS) runs it automatically ([details](#keeping-monitors-alive))
 - **Overload backoff** — detects sustained API overload (`429/500/502/503/504/529`) and retries on a configurable exponential backoff with jitter and a cumulative-wait cap, distinct from the usage-reset path ([details](#overload-backoff))
 - **Safeguard retry** — auto-continues past an AUP-safeguard false-positive (often transient), capped at a few tries so a sticky flag can't loop ([details](#safeguard-retry))
+- **Interrupted-stream resume** — picks the work back up when a laptop suspend or a dropped connection truncates a response mid-turn and leaves the session parked at an idle prompt ([details](#interrupted-stream-resume))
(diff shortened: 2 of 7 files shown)


cheapestinference (changes requested):
Good addition — the gap is real, the family shape is right (parallel bounded machine, shared `validateBoundedRetry`, idle-only, usage-limit precedence), the physical-suspend verification is exactly the standard I want for a new detection family, and the FP pins for mid-line quotes / user-typed lines / tool-echo / scrollback are all present and pass. Suite 552/0 here. Two exec-verified gaps before merge, both in `RENDER_HEAD`'s shape rule:

**1. An indented, glyph-less head is a quotation, and it fires.** `^\s*(?:[⏺●]\s+)?API Error:` admits arbitrary indentation with no glyph — but that shape is how a hanging-indent quotation renders. Both of these fire on this branch (verified against `acfc1b0`):

```
⏺ The error I saw before the crash was:
  API Error: Your computer went to sleep mid-response. The response above may be
  incomplete.
❯                                          → streamInterruptedMatch FIRES

❯ what does this mean?
  API Error: Connection lost mid-response. The response above may be incomplete.
❯                                          → FIRES
```

Idle prompt, healthy session, `isWorking` false, foreground fine → the monitor types `continue` twice. Bounded and mild, but it's the exact wrap/quotation class #75 closed for `SPEND_LIMIT`, and this PR's own doctrine states it: *a render starts at column 0 or behind a marker; indented with no marker is a wrap*. The real render's **head** is never indented — your verified capture has `⏺ API Error:` at column 0, and a narrow pane wraps the head's *tail* downward, never the head itself (it starts the line). So the split that fixed #75 applies verbatim:

```js
const RENDER_HEAD = /^(?:\s*[⏺●]\s+)?API Error:/i;   // glyphed head may indent; bare head is column 0
```

keeps both real forms (glyphed, and glyph-less print-mode/column-0) and kills both panes above. Please pin both as tests — the existing FP fixtures all carry the quote mid-line, which is why this class slipped through. (A quotation that reproduces the *glyph* at column 0 remains per-line indistinguishable — same boundary class as #74's limit-naming prose; a `KNOWN BOUNDARY` pin in the same style would be consistent.)

**2. The continuation scan walks across a block boundary.** In `streamInterruptedMatch`, a non-indented line inside the wrap window is `continue`d past instead of ending the render:

```
⏺ API Error: 529 {"type":"error","error":{"type":"overloaded_error"}}
⏺ I was explaining that your
  computer went to sleep mid-response and that truncates the turn.   → FIRES (verified)
```

The head is the 529 render; the phrase lives in the *next block's* indented continuation, two lines down. Monitor-level precedence (overload runs first) masks most real panes, so this is a correctness cleanup rather than an incident-in-waiting — but the fix is free: `break` out of the window on the first non-indented line after the head instead of `continue`.

Nothing else to flag. The config refactor is a net improvement (third copy avoided, overload correctly kept separate), the README/llms.txt coverage is thorough, and I checked the state machine against the safeguard branch line by line — deferral without counter reset, give-up latch, hold cooldown and foreground gating all match the house pattern. With the head anchored and the scan bounded, this merges.

nyxaria:
Both confirmed and fixed in `8a73ec8`. I reproduced each against `acfc1b0` before changing anything — your repros land exactly as described.

**1. Indented glyph-less head.** Taking the fix as proposed: the glyph and the indentation are one rule, not two.

```js
const RENDER_HEAD = /^(?:\s*[⏺●]\s+)?API Error:/i;   // glyphed head may indent; bare head is column 0
```

You're right that this path's own doctrine already stated the rule it was breaking — a real head *starts* its line, and a narrow pane wraps the head's tail downward, never the head itself. The FP fixtures all carried the quote mid-line, so the hanging-indent shape had no pin.

**2. Cross-block walk.** `break` on the first flush-left line after the head, instead of `continue`. Agreed on severity — overload precedence masks it on real panes, so correctness cleanup rather than an incident-in-waiting.

Verified after the patch — three FPs dead, all four real forms preserved:

```
inert  hanging-indent quote (model)          FIRES  real render (verified capture)
inert  hanging-indent quote (user)           FIRES  bare column-0 (print mode)
inert  cross-block walk                      FIRES  indented glyphed head
                                             FIRES  wrapped head/phrase split
```

**Five tests added**, all failing first: both quotation shapes, the cross-block walk, a bare column-0 head (the print-mode form the split exists to preserve — worth pinning, since the obvious fix of banning indentation outright would have killed it), and the `KNOWN BOUNDARY` pin in the #74 style for a quotation reproducing the glyph flush left. Suite 558/558.

On that boundary: it's the right call to accept it, and the cost is lower here than in the limit path — what fires is a bounded, self-clearing `continue`, not a multi-hour latched wait. Worst case the session gets one spurious "continue" while idle and the render scrolls away.

cheapestinference (approved):
Re-verified against 8a73ec8: both RENDER_HEAD repros from the review (hanging-indent quotation, user-question + indented quote) are now inert, the cross-block walk stops at the first flush-left line, and the real renders (glyphed col-0, glyphed indented TUI block, bare col-0 print-mode) still fire. Added checks on my side — mid-line quote, user-typed line, tool echo, stale render with work below — all inert. Suite 557/0. Merging.

PR #80, opened 2026-09-01
feat(launcher): name the auto-created tmux session (--tmux-session)

nyxaria:
## Problem

The auto-created session is named `claude-retry-<pid>-<timestamp>`. It's unique, but opaque — after a few launches `tmux ls` is a wall of timestamps with nothing to say which checkout each session belongs to, so re-attaching to a specific run means guessing at PIDs. There was no way to override it: the name was built inline in `createTmuxSession`.

## What this adds

```sh
claude --tmux-session api          # this launch
tmux attach -t '=api'              # ...and back to it later

export CLAUDE_AUTO_RETRY_SESSION_NAME=api   # every launch from this shell (direnv, per-project)
```

The flag wins over the env var and is **consumed by the launcher** — it's ours, not claude's, and claude would abort on an unknown option. Unnamed launches keep the generated name and behave exactly as before.

## Three details worth reviewing

- **`.` and `:` are normalized to `_` up front.** tmux reserves them as target separators and silently rewrites them while creating the session (verified on 3.5a — `new-session -s a.b` yields `a_b`, rc 0). Without normalizing, the name we hold stops matching the session tmux made and the follow-up `-t <name>` targets (`set-window-option`, `attach-session`) miss it. The launcher prints a line when it rewrites.
- **Attach now uses an exact-name target.** A chosen name can be a prefix of another session's (`api` vs `api-worker`) and tmux matches targets by prefix by default, which would attach to the wrong session. `-t '=api'` forces exact match. This also applies to the first-window target.
- **A duplicate name fails with something actionable.** Generated names never collided; a chosen one is the obvious way to trip. tmux's bare `duplicate session:` becomes a message naming the collision plus the `tmux attach` command for the existing session, still exit non-zero. It's a permanent failure, so it stays outside `retryTransientServerError` — the two patterns are disjoint, and there's a test pinning that.

Inside an existing tmux session, or in `--print` mode, no session is created and the flag does nothing — the launcher says so rather than ignoring it silently. The env var stays quiet in that case, since it rides along on every launch and would warn at every prompt.

## Testing

24 new unit tests in `test/launcher.test.js` (argv/env parsing, normalization, exact targeting, duplicate classification, the ignored-flag warning), plus one that runs the extracted argv through `buildTmuxInnerCmd` to pin that the flag never reaches the pane command. Full suite: **610 passing**.

Verified end-to-end against a real tmux server (isolated via `TMUX_TMPDIR`, stub `claude` on `PATH`):

| case | result |
|---|---|
| `--tmux-session my.api --model opus` | session `my_api` created; claude received exactly `--model opus` |
| same name again | duplicate message + attach command, exit 1 |
| `CLAUDE_AUTO_RETRY_SESSION_NAME=from-env` | session `from-env` |
| no name given | `claude-retry-<pid>-<ts>`, unchanged |
| `--tmux-session` with no value / empty value | usage error, exit 2 |
| `-p --tmux-session api` | flag consumed, "ignored" note, print mode runs |

README gets a "Naming the session" section; CHANGELOG entry under Unreleased.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   reset already in the past means the limit cleared — retry now rather than a year later.
 
 ### Added
+- **The auto-created tmux session can be named.** It was hardcoded to
+  `claude-retry-<pid>-<timestamp>`: unique, but opaque — `tmux ls` after a few launches is
+  a wall of timestamps and nothing says which checkout each session belongs to, so
+  re-attaching to a specific run meant guessing. `claude --tmux-session api` names one
+  launch and `CLAUDE_AUTO_RETRY_SESSION_NAME` names every launch from a shell; the flag
+  wins over the env var and is consumed by the launcher, so it never reaches `claude`
+  (which would reject it as an unknown option). Unnamed launches are unchanged. `.` and
+  `:` are normalized to `_` up front because tmux rewrites its own target separators as it
+  creates the session — without that the name we hold would stop matching the session tmux
+  made, and the follow-up `-t` targets would miss. A name already in use fails with the
+  `tmux attach` command for the existing session rather than a raw tmux error, and attach
+  now uses an exact-name target (`-t '=api'`) so a name that is a prefix of another
+  session's no longer resolves to the wrong one.
+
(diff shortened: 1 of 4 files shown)


nyxaria:
The usecase for me is having an orchestrator script which launches claude auto retry sessions when a event is detect (eg new issue or error in log) and autonomously fix the issue as it is detected.

